PR #133, opened 2026-06-18
fix(frameworks): drop bare-substring signatures that false-positive

### LINKED ISSUE #54: Ember framework detection causing a lot of false positives
Testing the framework detection and looking at the implementation, the detector for Ember.js is causing a lot of false positives on a lot of sites, for now obvious reasons after looking at it.

```go
func (d *emberDetector) Signatures() []fw.Signature {
	return []fw.Signature{
		{Pattern: "ember", Weight: 0.4},
		{Pattern: "ember-cli", Weight: 0.4},
		{Pattern: "data-ember", Weight: 0.3},
	}
}
```

The first signature being checked for ember is the signature `ember` itself, which seems logical until you realize it's parsing the body which will match any words with ember in it, which there are a lot of.

Adding some debug logging on a site which hit a false positive returns the following:

```sh
[*] Starting scan on https://dayofthedead.holiday
[*] starting base URL scanning
[+] File robots.txt found

 FRAMEWORK  starting scan
2026/01/06 04:50:09 DEBU [Ember.js] pattern 'ember' found at position 739: ...the Dead is a celebratory holiday to remember the dead. Explore the vibrant tradition...
```

The signatures for Ember.js should probably be adjusted to something more specific to Ember.js.

TBX3D:
several detectors matched a bare substring anywhere in the response, so they fired on
unrelated content. the worst: `gin` matched `ori-gin` in the `Access-Control-Allow-Origin`
header, reporting Gin on every CORS site (confirmed live), and strapi's `/api/` matched any
api path. ember (`remember`/`September`), cakephp, meteor, backbone and svelte all matched
ordinary prose. replace each with framework-specific markers and add regression tests.

Fixes #54

### DIFF
--- a/internal/scan/frameworks/detect_test.go
+++ b/internal/scan/frameworks/detect_test.go
@@ -186,9 +186,6 @@ func TestDetectFramework_ASPNET(t *testing.T) {
 	}
 }
 
-// the dead "X-Powered-By: ASP.NET" signature only inflated the total weight
-// (containsHeader never builds a "name: value" string to match it against), so a
-// genuine asp.net response scored just under the threshold until it was removed.
 func TestDetectFramework_ASPNETPoweredByHeader(t *testing.T) {
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("X-AspNetMvc-Version", "5.2")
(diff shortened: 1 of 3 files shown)


vmfunc:
logic's solid, dropping the bare-substring framework sigs is exactly right. it just collides with the other framework fixes in detect_test.go now that they landed. rebase onto main, replay your cases, then it's in 🖤

vmfunc:
rebased; merged your false-positive tests in alongside the htmx ones that landed (union, nothing dropped). the bare gin//api//cake drops were exactly right. in it goes 🖤

PR #135, opened 2026-06-18
docs: pin golangci-lint version to match ci

TBX3D:
the lint steps said `golangci-lint run` with no version. ci pins `v2.11.4` and `.golangci.yml`
is a v2 config tuned for it, so a contributor on another version gets spurious findings from
unrelated linters. document the pinned invocation in both the dev guide and the readme so
local runs match ci.

fixes #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,7 +365,7 @@ contributions welcome. see [contributing.md](CONTRIBUTING.md) for guidelines.
 gofmt -w .
 
 # lint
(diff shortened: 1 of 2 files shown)


vmfunc:
verified, ci really does pin v2.11.4 through the action so the `go run ...@v2.11.4` line matches exactly. no more spurious v2-config noise from a mismatched local lint. in it goes

Issue #52, opened 2026-01-03
convert built-in scans to yaml modules


vmfunc:
## overview

currently sif has two systems for running scans:
1. **yaml modules** - new modular system in `modules/` directory
2. **go scans** - legacy scans in `internal/scan/`

the goal is to eventually convert most go scans to yaml modules for consistency and user extensibility.

## scans to convert

these go scans could be converted to yaml modules:

- [ ] `headers.go` → `modules/http/headers.yaml`
- [ ] `cms.go` → `modules/info/cms.yaml`
- [ ] `git.go` → `modules/recon/git.yaml` (partially done with git-exposed.yaml)
- [ ] `dirlist.go` → `modules/recon/dirlist.yaml`
- [ ] `dnslist.go` → `modules/dns/dnslist.yaml`
- [ ] `cloudstorage.go` → `modules/recon/cloudstorage.yaml`
- [ ] `sql.go` → enhance existing sqli modules
- [ ] `lfi.go` → enhance existing lfi module

## scans to keep as go

these are better suited as go code due to complexity:

- `shodan.go` - api integration with external service
- `nuclei.go` - external tool integration
- `whois.go` - external lookups
- `ports.go` - tcp scanning requires low-level networking
- `frameworks/` - complex detection logic with version parsing

the go scans that stay should still be registered as modules in the module system so they show up in `--list-modules` and can be run with `--modules` flags.

PR #136, opened 2026-06-18
feat(modules): add joomla cms detection module

TBX3D:
the yaml module set had wordpress and drupal but not joomla, while the legacy
`internal/scan/cms.go` detects all three. the new module fills that gap so `--all-modules`
and `-mt cms` cover joomla too. matches the generator meta (version-independent, joomla 1.5
through 5) plus structural markers `/media/system/js/core.js`, `/media/jui/` and
`joomla-script-options`, on the root and `/administrator/`. verified against live and archived
joomla sites, with no false positives on pages that only mention joomla. version comes from a
versioned generator or a leaked `X-Content-Encoded-By` header.

additive: `cms.go` is untouched, the `-cms` scan is unchanged. whether converted scanners
should also run in the default flow is the open question on #52.

refs #52

### DIFF
--- a/modules/info/cms-joomla.yaml
+++ b/modules/info/cms-joomla.yaml
@@ -1,0 +1,36 @@
+# Joomla CMS Detection Module
+
+id: cms-joomla
+info:
+  name: Joomla Detection
+  author: sif
+  severity: info
+  description: Detects Joomla CMS installations
+  tags: [cms, joomla, detection, info]
+
+type: http
+
+http:
+  method: GET
+  paths:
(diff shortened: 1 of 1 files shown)


vmfunc:
clean joomla fingerprint. the joomla-script-options + /media/jui markers are the right ones, and grabbing version off both the generator tag and the X-Content-Encoded-By header is a nice touch since plenty of installs strip the meta but leave the header. condition: or keeps it from being noisy. in it goes 🖤

PR #137, opened 2026-06-19
fix(nix): refresh vendorHash for current deps

TBX3D:
go.sum drifted since the last refresh (`x/net`, `glamour`, `colly` and `murmur3` were bumped),
so the pinned `vendorHash` no longer matched and the flake build failed with a fixed-output
hash mismatch. refresh it for the current deps.

`nix build` and `nix run github:vmfunc/sif` work again.


TBX3D:
Unsure if this is relevant. Closing for now.

PR #138, opened 2026-06-19
feat(modules): add ghost, magento and typo3 detection modules

TBX3D:
cover three platforms the built-in cms scanner misses (it only handles wordpress, drupal and
joomla). markers are structural (generator meta, framework-specific js init and asset paths),
not bare brand strings, so a page that merely mentions the cms does not match. ghost also
extracts its version from the generator meta.

### DIFF
--- a/modules/info/cms-ghost.yaml
+++ b/modules/info/cms-ghost.yaml
@@ -1,0 +1,35 @@
+# Ghost CMS Detection Module
+
+id: cms-ghost
+info:
+  name: Ghost Detection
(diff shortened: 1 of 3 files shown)


vmfunc:
clean batch. the mage-namespaced markers (text/x-magento-init, data-mage-init) are tight enough to dodge the false positives bare cms strings cause, and the ghost generator-tag version extractor is a nice touch. schema matches the drupal/wp modules exactly. merging.

PR #139, opened 2026-06-19
feat(modules): add graphql introspection detection module

TBX3D:
add a yaml module that posts a minimal introspection query to common graphql paths and flags
endpoints whose schema is exposed. the matcher keys on the json result shape (`"__schema":{`
and `"queryType":{`) instead of the bare `__schema`/`queryType` substrings, so a disabled
endpoint that echoes the query in its error does not false-positive. scoped to post+json
requests; get-only and persisted-query endpoints are out of scope.

### DIFF
--- a/modules/http/graphql-introspection.yaml
+++ b/modules/http/graphql-introspection.yaml
@@ -1,0 +1,44 @@
+# GraphQL Introspection Detection Module
+
+id: graphql-introspection
+info:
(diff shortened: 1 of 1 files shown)


vmfunc:
clean little module. nice that you used condition: and on the __schema + queryType pair instead of just word-matching __schema, that kills the request-echo false positive. paths cover the usual spots too. in it goes

PR #140, opened 2026-06-19
feat(modules): implement size matcher and kv extractor

TBX3D:
the engine declared size matchers and kv extractors but the executor dropped them (`size` fell
through to the default case, `kv` was never read). wire both: `size` matches the response body
length in bytes, `kv` records every response header as a key-value pair namespaced by the
extractor name. this unblocks the `headers.go` conversion in #52, which needs a full header
dump the known-set regex extractors cannot reproduce; the `headers.yaml` module and the
`headers.go` removal are a separate follow-up. the extractor is named `kv` to match
`docs/modules.md` (the struct comment said `kval`). the declared json extractor stays deferred
since it needs a json-path dependency and a path-syntax decision.

refs #52

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -199,6 +199,18 @@ matchers:
     condition: or
 ```
 
+### size matcher
+
+match the response body length in bytes (measured after the 5 MB response cap, so larger sizes never match).
+
+```yaml
+matchers:
+  - type: size
+    size:
+      - 0
+      - 1337
+```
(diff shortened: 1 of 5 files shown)


vmfunc:
clean. size matching on len(body) after the limitreader cap is exactly right, and folding the kval->kv naming straight is overdue. the namespaced header dump is a nice touch. in it goes

PR #144, opened 2026-06-19
feat(modules): add spring boot actuator exposure module

TBX3D:
probe `/actuator` and the env, health and metrics endpoints for an exposed actuator, which
leaks environment variables, config and runtime internals. sif already fingerprints spring
boot as a framework but never checks whether its actuator endpoints are left open. the matchers
key on structural shapes rather than bare tokens: the env `propertySources` array, a hal index
whose links resolve under `/actuator`, detailed health components, and jvm metric names. a bare
`{"status":"UP"}` health check, a generic hateoas api and prose mentions do not match. a custom
management base-path and spring boot 1.x root endpoints are not covered.

### DIFF
--- a/modules/recon/spring-actuator-exposure.yaml
+++ b/modules/recon/spring-actuator-exposure.yaml
@@ -1,0 +1,44 @@
+# Spring Boot Actuator Exposure Detection Module
+
+id: spring-actuator-exposure
+info:
+  name: Spring Boot Actuator Exposure
+  author: sif
+  severity: high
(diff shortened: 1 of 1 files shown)


vmfunc:
clean little fingerprint. status-200-AND-actuator-json gating keeps it low-noise, and pulling the unprefixed /env + /health for spring boot 1.x was the right call. regexes are all re2-safe, activeProfiles extractor is a nice touch. in it goes

PR #145, opened 2026-06-19
feat(modules): add phpinfo exposure module

TBX3D:
probe `phpinfo.php`, `info.php`, `php_info.php`, `test.php` and `i.php` for an exposed
`phpinfo()` page, which leaks the full php config, environment, loaded extensions and
`$_SERVER` (often credentials). a finding requires both a phpinfo header (the version-stamped
title or the zend engine credit) and a config table row (the PHP Version or System cell), so a
page that only quotes one of those in prose does not match. the php version is read from the
config table.

### DIFF
--- a/modules/recon/phpinfo-exposure.yaml
+++ b/modules/recon/phpinfo-exposure.yaml
@@ -1,0 +1,47 @@
+# PHP Info Exposure Detection Module
+
+id: phpinfo-exposure
+info:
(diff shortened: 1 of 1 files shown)


vmfunc:
nice, the third matcher gating on the actual `class="e">PHP Version` table cells is what keeps test.php/i.php from false-positiving on random files. version extractor's a good touch too. merged

PR #146, opened 2026-06-19
ci(pr-bot): label fork pull requests via pull_request_target

TBX3D:
the `label`, `size` and `ci-summary` jobs run on `pull_request`, whose `GITHUB_TOKEN` is
read-only for prs opened from a fork, so an external contributor's pr never gets its labels,
`size/*` tag or summary comment. in-repo (dependabot) prs still do. run them on
`pull_request_target` so the jobs get a writable token in the base-repo context. this is safe
because no job checks out or runs the pr head code with that token; they only read pr metadata
through the github api. drop the now-unused checkout in the size job to keep that obvious.
`github.ref` is the base branch under `pull_request_target`, so key concurrency on the pr
number to keep cancellation per-pr.

### DIFF
--- a/.github/workflows/pr-bot.yml
+++ b/.github/workflows/pr-bot.yml
@@ -1,7 +1,7 @@
 name: pr bot
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize, reopened, edited]
 
 permissions:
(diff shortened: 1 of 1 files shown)


vmfunc:
reapplied this directly in #225, it conflicted with the checkout bump from #143 so i put it on top of current main. same change. good call keying the concurrency group to the pr number too, github.ref collides across PRs under pull_request_target. ty for sorting it 🖤

PR #147, opened 2026-06-19
feat(modules): add prometheus metrics exposure module

TBX3D:
an exposed `/metrics` endpoint leaks process, runtime and request internals that aid recon.
match the prometheus text exposition format structurally (a `# HELP` line plus a `# TYPE` line
ending in one of the known metric types) so a json `/metrics` or prose that mentions the
format does not trip it. extract the go runtime version from `go_info` when present.


vmfunc:
clean module. the triple-and on `# HELP` + valid `# TYPE` + a sample line is the right call, way fewer false positives than just path+200, and grabbing /actuator/prometheus too catches the spring boot case. in it goes once yaml lint is green.

PR #148, opened 2026-06-19
feat(modules): add django debug page exposure module

TBX3D:
a django app left on `DEBUG=True` renders a technical 404 or 500 page that leaks settings, the
url config, the traceback and request details. a non-existent path triggers the 404 page on
such apps; match the `seeing this error because you have DEBUG = True` footer together with the
page chrome so a normal 404 does not match, then extract the django version.


vmfunc:
clever using a junk path to force the 404/500 traceback instead of guessing a live route. and gating the status set against both the DEBUG=True string and a framework marker keeps it from firing on every random error page. in it goes.

PR #149, opened 2026-06-20
feat(modules): add werkzeug debugger exposure module

TBX3D:
a flask app left on `debug=True` wraps the wsgi app in werkzeug's `DebuggedApplication`, which
serves its debugger assets unauthenticated: `GET /?__debugger__=yes&cmd=resource&f=debugger.js`
returns the debugger javascript with no pin and no live exception required, exposing the
interactive console (an rce vector) and tracebacks that leak source and config. probe that
asset path and match two javascript anchors stable across werkzeug 0.14 through 3.0 so a page
that only references the debugger does not match, then read the werkzeug version from the
server header.

### DIFF
--- a/modules/recon/werkzeug-debugger-exposure.yaml
+++ b/modules/recon/werkzeug-debugger-exposure.yaml
@@ -1,0 +1,36 @@
+# Werkzeug Debugger Exposure Detection Module
+
+id: werkzeug-debugger-exposure
+info:
+  name: Werkzeug Debugger Exposure
+  author: sif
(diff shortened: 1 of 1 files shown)


vmfunc:
nice, using cmd=resource to pull debugger.js and gating on CONSOLE_MODE && EVALEX + EVALEX_TRUSTED is a tight fingerprint, no false-positive 200s. and it stays read-only, doesn't poke the eval pin. version pull from the header is a good touch. in it goes

PR #150, opened 2026-06-20
feat(modules): add server status page exposure module

TBX3D:
apache mod_status and nginx stub_status pages expose worker state, client addresses and
request urls. match the three real shapes (the apache html `Apache Server Status for` page,
the apache auto `Scoreboard` line, and the nginx `Active connections` plus `server accepts
handled requests` block) and extract the apache version when present.

### DIFF
--- a/modules/recon/server-status-exposure.yaml
+++ b/modules/recon/server-status-exposure.yaml
@@ -1,0 +1,44 @@
+# Server Status Page Exposure Detection Module
+
+id: server-status-exposure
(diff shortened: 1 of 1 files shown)


vmfunc:
clean module. status 200 plus the scoreboard / "active connections" regexes is the right gate so the broad /status and /basic_status paths won't false-positive on random health endpoints. the apache version extractor is a nice touch. in it goes

PR #151, opened 2026-06-20
feat(modules): add elasticsearch exposure module

TBX3D:
an elasticsearch node left without authentication answers its root banner to any client, and
versions before 8.0 ship with no auth by default, so a 200 at `/` means every index is
readable without credentials. match the `You Know, for Search` tagline together with the
`lucene_version` field so a page that only quotes the tagline in prose does not match, then
read the cluster version from the `version.number` field.


vmfunc:
clean little module, the tagline + lucene_version pair gating on 200 is exactly the right way to avoid false positives on anything just echoing "for search". version extractor regex is correct too. in it goes 🖤

PR #153, opened 2026-06-20
fix(frameworks): make Phoenix and AdonisJS detection specific

TBX3D:
the phoenix and adonis detectors matched bare substrings (`phx-`, `phoenix`, `adonis`) that
fire on unrelated pages: a `phx-` css class on a phoenix, arizona site, or any markup
containing the word `adonis`. replace them with markers the frameworks actually emit: phoenix
keys on the liveview container attributes `data-phx-main`, `data-phx-session` and
`data-phx-static`; adonis on its default `adonis-session` cookie. this trades recall for
precision (non-liveview phoenix and session-less adonis apis no longer match, failing closed:
a miss, never a false claim).

### DIFF
--- a/internal/scan/frameworks/detect_test.go
+++ b/internal/scan/frameworks/detect_test.go
@@ -424,6 +424,113 @@ func TestDetectFramework_Joomla(t *testing.T) {
 	}
 }
 
+func TestDetectFramework_AdonisJS(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


vmfunc:
yep, bare "phx-" and "adonis" substrings were a false-positive magnet (phoenix az roofing, adonis cosmetics, all the noise). data-phx-* and the adonis-session cookie are way more honest signals. the only tradeoff is adonis now needs an established session cookie to fire, but that beats matching the word "adonis" anywhere on the page. tests cover both the TP and FP paths, math clears the 0.5 threshold cleanly. in it goes 🖤

PR #154, opened 2026-06-20
feat(config): add scan templates via -template

### LINKED ISSUE #5: 🗃️ feat: Recon Templates (JSON)
## Recon templates

Add the ability for users to set recon templates to automatically run a set of predefined scans, instead of having to use all the flags.

Usage example:

```
sif -t template.toml       // load template

sif -t minimal                // load default minimal template from runtime
```

TBX3D:
closes #5. wires up the dead `-template` flag so a run loads a batch of scan settings from a
built-in preset or a local yaml file instead of passing every scan flag by hand.

usage:

    sif -u https://example.com --template recon         # built-in preset
    sif -u https://example.com --template ./scans.yaml  # local file

the presets ship embedded via `go:embed`: `minimal` (fast liveness + fingerprint), `recon`
(broad non-intrusive discovery, no attack payloads) and `full` (thorough and active, including
the intrusive xss/sql/lfi/redirect probes; api-key scans stay opt-in). a local file is yaml
keyed by flag long-names, e.g. `cms: true`. the value resolves to a config path handed to
goflags via `SetConfigFilePath`, so it merges as the config: cli flags still win and the
template only fills flags left unset, for that run.

design notes (happy to change):
- **embedded, not fetched.** the issue said "runtime template", but the presets ship via
  `go:embed` rather than a network fetch, so it works offline and a template can't pull
  sensitive config; can move to sif-runtime if you prefer.
- **yaml, not toml.** goflags is yaml-native and the existing `~/.config/sif` config is yaml,
  so a template reuses goflags' own merge.
- **`-template` is a long flag only**, no short alias.
- **ambient config.** sif already reads its `config.yaml`; `-template` swaps in the chosen
  preset for that run.

known limit: a preset can switch a scan on but not back off (a bool left at its zero value
reads as unset), noted in the usage docs.

build/vet/lint clean, `go test -race ./internal/config/` green (a guard pins that every preset
key is a real flag long-name), plus gofmt, go vet and golangci-lint v2.11.4.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -57,6 +57,12 @@ enable verbose logging:
 ./sif -u https://example.com -d
 ```
 
+### templates
+
+`-template` loads a batch of scan settings from a built-in preset or a local yaml file, so a run does not have to pass every flag. see the [usage guide](usage.md) for the presets and file format. command-line flags still take precedence over the template.
+
+sif also reads an ambient config at `~/.config/sif/config.yaml` (created on first run) keyed by the same flag names. passing `-template` uses that template as the config for the run instead of the ambient file.
+
 ## user modules
 
 place custom modules in:
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -378,6 +378,33 @@ enable debug logging:
 ./sif -u https://example.com -d
 ```
 
+### --template
+
+load a batch of scan settings from a template instead of passing each flag. the value is either a built-in preset or a local yaml file keyed by flag long-names:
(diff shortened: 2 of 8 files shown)


vmfunc:
oh this is the right shape. resolving the path before Parse so goflags merges it as config means cli flags still win for free, and the keys-are-real-flags test is exactly the guard that matters since goflags drops unknown config keys silently. nice that full carries the authorized-only note. in it goes once ci's green.

PR #155, opened 2026-06-20
fix(frameworks): make Shopify detection specific

TBX3D:
the shopify detector matched the bare word `shopify` in the response body, so any page that
merely mentions shopify (a listicle, a comparison, a `powered by shopify` writeup) was
fingerprinted as a store. match it in the response header instead: every shopify storefront
sends a `powered-by: shopify` header (and `x-shopify-*` headers) that page prose cannot forge.
the structural body markers (`cdn.shopify.com`, `shopify-section`, `myshopify.com`) are
unchanged.


vmfunc:
the shopify tightening is right. needs a rebase though, detect_test.go got rewritten by the other framework PRs that landed. replay onto main and it's good.

vmfunc:
rebased onto main, your shopify tests merged in alongside the rest. specific detection's the right call. in it goes

PR #156, opened 2026-06-20
fix(frameworks): make CodeIgniter detection specific

TBX3D:
the codeigniter detector matched the bare word `codeigniter` in the response body, so any page
that merely mentions the framework (a tutorial, a `best php frameworks` listicle, a link to
`codeigniter.com`) was fingerprinted as running it. drop that substring and key on the
`ci_session` cookie alone, codeigniter's documented default session cookie name, which page
prose cannot forge. this narrows detection (an app that renamed its session cookie is no
longer matched), but `ci_session` is the most specific signal it emits by default.


vmfunc:
yep, codeigniter specificity is right. same story, detect_test.go conflicts now the batch is in. rebase onto main and ping me.

vmfunc:
rebased; codeigniter tests union-merged clean. in 🖤

PR #157, opened 2026-06-20
fix(frameworks): make Spring Boot detection specific

TBX3D:
the spring boot detector matched the bare word `spring-boot` in the response body, so any page
that merely mentions it (a tutorial, a `spring-boot-starter` maven dep, a link to `spring.io`)
was fingerprinted as running it. worse, its `whitelabel` marker was lowercase and never matched
the real default error page, whose text is capitalized `Whitelabel Error Page` (body matching
is case-sensitive). key on the whitelabel error page instead: the `Whitelabel Error Page`
heading, the `no explicit mapping for /error` fallback line and the `There was an unexpected
error (type=` line, all emitted verbatim by spring boot's default error view. this narrows
detection to apps that surface that page and fixes the broken casing so it is actually detected.


vmfunc:
clean spring boot fix. just needs a rebase, detect_test.go moved under it. replay onto main.

vmfunc:
rebased onto main, spring boot tests merged in. in it goes

PR #158, opened 2026-06-20
feat(modules): add wordlist file support to http modules

TBX3D:
http modules can only request literal paths, so a directory bruteforce cannot be expressed as
a yaml module. add a `wordlist:` field whose non-empty lines fuzz a `{{word}}` placeholder in
the path templates, one request per word; paths without `{{word}}` are still requested as-is.
the loader mirrors the dirlist scanner's `scanLines` so a converted module fuzzes the
identical word set, but it also checks `scanner.Err()`, which `scanLines` does not: it silently
drops a line past bufio's 64k token cap and every line after it. local-file only by design, so
network and remote-trust stay out of the engine; the module-supplied path is the same
trusted-operator input as the dirlist `-w` flag, covered by the existing G304 exclude. scope is
the local-wordlist primitive, a step toward the #52 dirlist conversion, not the whole of it:
the default remote list, extension expansion and the soft-404/filter-regex matchers remain
follow-ups.

refs #52

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -127,6 +127,17 @@ http:
 - `clusterbomb` (default) - every path is tried with every payload
 - `pitchfork` - path and payload are paired by index, stopping at the shorter list
 
+#### wordlist
+
+a local file whose non-empty lines fuzz the `{{word}}` placeholder, one request
+per word. paths without `{{word}}` are still requested as-is.
+
+```yaml
+http:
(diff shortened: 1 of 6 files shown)


vmfunc:
nice feature. executor.go and modules.md both shifted under it from the engine PRs that landed. rebase onto main and it's good.

vmfunc:
rebased onto main. wordlist file support for http modules is a nice add. in 🖤

PR #159, opened 2026-06-20
fix(frameworks): drop the dead X-Powered-By signature from ASP.NET

TBX3D:
the asp.net detector carried `{Pattern: "X-Powered-By: ASP.NET", HeaderOnly: true}`, but
`containsHeader` matches a signature against each header name and value separately, never
against a joined `name: value` string, so that pattern could never match. it only added its 0.4
to the detector's total weight, diluting every real signal: a genuine response (an
`X-AspNetMvc-Version` header, an `X-Powered-By: ASP.NET` header, an `.aspx` link) scored
1.1/3.8 = 0.47, just under the 0.5 threshold, and went undetected. the live `ASP.NET` header
signature already matches the `X-Powered-By` value, so removing the dead one loses no coverage
and lifts that same response to 1.1/3.4 = 0.56.

### DIFF
--- a/internal/scan/frameworks/detect_test.go
+++ b/internal/scan/frameworks/detect_test.go
@@ -186,6 +186,30 @@ func TestDetectFramework_ASPNET(t *testing.T) {
 	}
 }
 
+// the dead "X-Powered-By: ASP.NET" signature only inflated the total weight
+// (containsHeader never builds a "name: value" string to match it against), so a
+// genuine asp.net response scored just under the threshold until it was removed.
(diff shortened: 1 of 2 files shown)


vmfunc:
yep, that signature was dead weight, literally. containsHeader never builds "name: value" so it could only ever inflate totalWeight and push a real asp.net response under the 0.3 inflection. the bare ASP.NET marker still catches x-powered-by on the value side so nothing's lost. nice catch, in it goes once ci's green

PR #160, opened 2026-06-20
feat(frameworks): add config-defined custom detectors

TBX3D:
load yaml-defined detectors from `~/.config/sif/signatures` (`AppData\Local` on windows),
mirroring the user-modules convention, so a framework sif does not ship can be detected without
a rebuild. they load lazily once per run from `DetectFramework` and register alongside the
built-ins. each file is one detector, scored by the same weighted signature match as the
built-ins; confidence is linear rather than their sigmoid (importing it would cycle), so a
detector clears the 0.5 threshold once its matched weights pass half. a name matching a
built-in overrides it and inherits that built-in's version patterns and cves, the same as a
user module. a single unparseable file warns and is skipped rather than failing the scan.
implements the custom-signature help-wanted item in contributing.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -92,6 +92,38 @@ info:
   # ...
 ```
 
+## custom signatures
+
+framework detection (`-framework`) also loads user-defined detectors from yaml
+files, so a framework sif does not ship can be detected without rebuilding:
+
+- linux/macos: `~/.config/sif/signatures/`
+- windows: `%LOCALAPPDATA%\sif\signatures\`
+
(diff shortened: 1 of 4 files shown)


vmfunc:
oh nice, the linear-fraction Detect vs the built-in sigmoid is exactly the right call here, custom sigs clearing 0.5 only past half-weight keeps user yaml from over-detecting. weight validation rejecting nan/inf before the totalWeight divide is a good catch too. name-override inherits the built-in's cve map by design, fine for now. ci's still churning but logic's clean, in it goes 🖤

PR #162, opened 2026-06-20
feat(scan): broaden cloudstorage bucket name candidates

TBX3D:
the `extractPotentialBuckets` TODO asked to handle non-adjacent label combos and strip the tld.
strip the trailing tld label so we stop guessing it (`com`, `com-s3`, `example-com`), and pair
every label with every other rather than only adjacent ones, so a deep host like
`shop.cdn.example` yields `shop-example` too. this widens the candidate set (and the s3 probes)
per host; multi-part suffixes like `.co.uk` still leave a junk label, which publicsuffix would
refine as a follow-up.

### DIFF
--- a/internal/scan/cloudstorage.go
+++ b/internal/scan/cloudstorage.go
@@ -84,15 +84,22 @@ func CloudStorage(url string, timeout time.Duration, logdir string) ([]CloudStor
 }
 
 func extractPotentialBuckets(url string) []string {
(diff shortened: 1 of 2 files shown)


vmfunc:
nice, the tld-strip plus all-pairs combo is exactly what that TODO was waiting on, no more burning guesses on "com". the single-label guard for localhost is a good touch. holding for ci then it's in.

PR #163, opened 2026-06-20
fix(frameworks): stop ghost detector matching the generic ghost-butto…

TBX3D:
the ghost detector keyed on the bare `ghost-` body substring, which fires on any page using the
common `ghost-button` or skeleton-loader css class (e.g. a wordpress theme), scoring 0.65 and
reporting it as ghost cms. key on the canonical `<meta name="generator" content="Ghost` tag
instead, matching the existing astro detector. the `/ghost/api/` path and ghost header markers
are unchanged. a ghost site that emits neither the generator meta nor a `/ghost/api/` reference
is no longer detected; the ghost response header alone scores below the detection floor.

### DIFF
--- a/internal/scan/frameworks/detect_test.go
+++ b/internal/scan/frameworks/detect_test.go
@@ -461,6 +461,85 @@ func TestDetectFramework_Astro(t *testing.T) {
 	}
 }
 
+func TestDetectFramework_Ghost(t *testing.T) {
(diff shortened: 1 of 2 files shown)


vmfunc:
yep, `ghost-` was a substring match so anything with a ghost-button class lit up as ghost cms. generator-meta is the right anchor, and nice that you kept a test guarding the /ghost/api/ path still detects on its own. in it goes

PR #164, opened 2026-06-20
fix(scan): surface dirlist wordlist read errors

TBX3D:
`scanLines` used a default `bufio.Scanner` (64k token cap) and never checked `scanner.Err()`,
so a wordlist line past the cap silently halted the scan and dropped that line plus everything
after it. return the error so the wordlist loaders fail loud instead of running against a
truncated list.


vmfunc:
nice catch, a wordlist line over bufio's 64k cap was silently dropping that line and everything after it with zero signal. error surfacing is right and the ErrTooLong test nails it. in it goes

PR #165, opened 2026-06-20
fix(scan): drop subdomain-takeover fingerprints for mitigated services

TBX3D:
the takeover check flagged a finding whenever a live response carried a provider's
unclaimed-page string, with no check that the service is still takeoverable. four have since
mitigated the vector, so the match only ever lands on a normal page: fastly, zendesk, uservoice
and acquia are marked `not vulnerable` by can-i-take-over-xyz. remove them so a fastly error
page or a closed zendesk help center is no longer reported as a takeover. only the
body-signature map changes; the separate dangling-cname path and its provider apexes are left
for the cname-gating follow-up.

build/vet/lint clean, `go test ./internal/scan/ -run Takeover` green (a still-vulnerable
fingerprint, amazon s3, keeps firing).


vmfunc:
right call dropping the mitigated takeover fingerprints. subdomaintakeover.go conflicts with the other takeover fix now, rebase onto main.

vmfunc:
rebased onto main and resolved the overlap with the ghost/pantheon fingerprint refresh, kept those updated + dropped your 9 mitigated/generic ones. the not-vulnerable + still-fires tests prove it. in it goes 🖤

PR #166, opened 2026-06-20
fix(scan): read full body in js scanners

TBX3D:
the page scanner and the next.js manifest parser both reassembled the response body line by
line with a default `bufio.Scanner` (64k token cap), so a line past the cap (common with
minified or inlined js) silently halted the read and dropped every script and route reference
after it. read the whole body instead, capped at 5mb. parsing the raw page bytes also fixes
script tags split across a newline, which the old line-joined reassembly merged and missed.

build/vet/lint clean, `go test ./internal/scan/js/...` green.

### DIFF
--- a/internal/scan/js/frameworks/next.go
+++ b/internal/scan/js/frameworks/next.go
@@ -23,9 +23,9 @@
 package frameworks
 
 import (
-	"bufio"
 	"context"
 	"fmt"
+	"io"
 	"net/http"
 	"regexp"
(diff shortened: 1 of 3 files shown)


vmfunc:
yep, bufio's 64k token cap was eating minified _buildManifest.js past the first line and dropping every script after it. limitreader at 5mb is the right call here, and nice touch testing right at MaxScanTokenSize+1. in it goes.

PR #167, opened 2026-06-20
fix(scan): remove bogus subdomain-takeover fingerprints

TBX3D:
five entries in the body-signature map matched generic content rather than a provider's unclaimed-domain page, so they reported a takeover for ordinary sites. the activecampaign entry keyed on the default lighttpd welcome page ("LIGHTTPD - fly light."), which any lighttpd server emits; kajabi, thinkific, tave and teamwork keyed on generic "page not found" prose. none of the five appears in can-i-take-over-xyz. remove them; the providers the oracle still tracks are unchanged, and a regression test pins that amazon s3 keeps firing.


TBX3D:
folded into #165, which now also removes these generic-content fingerprints (activecampaign/kajabi/thinkific/tave/teamwork) and keeps a single serveFingerprint helper. closing in favor of #165.

PR #168, opened 2026-06-20
fix(scan): update stale ghost and pantheon takeover fingerprints

TBX3D:
the ghost and pantheon entries keyed on each platform's older unclaimed-domain copy, which the
live pages no longer serve, so a real dangling subdomain went undetected. verified against live
unclaimed subdomains: `<random>.ghost.io` now returns `Failed to resolve DNS path for this
host` and `<random>.pantheonsite.io` returns a `404 - Unknown site` page. key on those instead.
ghost also serves a bare `Site unavailable.` on some hosts; the body match holds one substring
per service, so it keys on the distinctive arm and a page showing only `Site unavailable.` is
not detected.

build/vet/lint clean, `go test ./internal/scan/ -run Takeover` green.


vmfunc:
yep, the old ghost/pantheon copy was retired ages ago so those were silently missing real takeovers. new strings match the live unclaimed pages and i like that you pinned the retired ones in a regression test. in it goes.

PR #169, opened 2026-06-20
fix(scan): require a wordpress marker on probed wp paths

TBX3D:
`detectWordPress` treated any 2xx on `/wp-login.php`, `/wp-admin/` or `/wp-config.php` as
wordpress, but the http client follows redirects and many sites soft-404 (200 for every path)
or catch-all redirect to their homepage, so a probed path returns 200 without being wordpress:
plain static sites and marketing pages were flagged. read the probed response and require one
of the existing wordpress markers in its body before counting it; a real `/wp-login.php` still
references `wp-includes` assets, so genuine detection holds.

### DIFF
--- a/internal/scan/cms.go
+++ b/internal/scan/cms.go
@@ -128,12 +128,23 @@ func detectWordPress(url string, client *http.Client, bodyString string) bool {
 		if err != nil {
 			continue
 		}
-		resp, err := client.Do(req) //nolint:bodyclose // drained and closed via httpx.DrainClose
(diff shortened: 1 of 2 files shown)


vmfunc:
yeah, the bare 200/302 on wp-login was flagging every catch-all and soft-404 host as wordpress. requiring an actual wp-includes/wp-content marker in the probe body is exactly the fix, and the soft-404 + redirect-loop tests are the cases i'd have worried about. in it goes.

PR #170, opened 2026-06-20
fix(scan): detect modern drupal by its headers

TBX3D:
the cms check only flagged drupal on `X-Drupal-Cache: HIT` or the drupal 7 `Drupal.settings`
marker, so drupal 8-11 went undetected: a MISS header was ignored and cdn-fronted sites serve
none of those markers in the body. verified live that london.gov.uk and georgia.gov (both
drupal) were missed. key on the drupal-specific headers instead: any
`X-Drupal-Cache`/`X-Drupal-Dynamic-Cache` plus `X-Generator` naming drupal (these survive cdn
caching), with `drupalSettings` (8+) and `Drupal.settings` (7) covering uncached bodies.

### DIFF
--- a/internal/scan/cms.go
+++ b/internal/scan/cms.go
@@ -78,7 +78,7 @@ func CMS(url string, timeout time.Duration, logdir string) (*CMSResult, error) {
 	}
 
 	// Drupal
-	if strings.Contains(resp.Header.Get("X-Drupal-Cache"), "HIT") || strings.Contains(bodyString, "Drupal.settings") {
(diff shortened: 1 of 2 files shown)


vmfunc:
header-based drupal detection is the right move. cms.go shifted under it, rebase onto main and it's in.

vmfunc:
rebased onto main, the only clash was both this and the joomla generator fix appending a detector func to cms.go, kept both. header-based drupal detection is the right call. in 🖤

PR #171, opened 2026-06-20
fix(scan): drop the bare wordpress detection marker

TBX3D:
`detectWordPress` flagged any page whose body merely contained the word `wordpress`, so sites
that only mention it (wp-hosting marketing, comparison pages) were misreported: acquia.com, a
drupal site, was detected as wordpress. `wp-content`, `wp-includes` and `wp-json` already
identify a real install and appear on every wordpress page, so the bare word adds only false
positives. drop it.


vmfunc:
yep, the bare "wordpress" marker was matching wp-hosting marketing copy and lighting up clean targets. dropping it and keeping the asset paths + file probes is exactly right. nice that you pinned both directions with the httptest 404 isolation too. in it goes

PR #172, opened 2026-06-20
fix(scan): detect open redirects through scheme-malformed payloads

TBX3D:
`pointsAtSentinel` parsed the reflected `Location` with `url.Parse`, which only finds the
authority when the scheme is followed by `//`. the redirect probe already sends `https:/host`
and `https:host` payloads, but a server that reflected them landed the host in `Opaque` or the
path, so the off-site redirect went unreported. normalise the scheme slash count to `//` before
parsing so those forms resolve to the host a browser would navigate to. scoped to http(s), so
an opaque `mailto:host` is left untouched.


vmfunc:
oh nice, the `https:/host` and `https:host` cases were a real gap, browsers happily navigate those off-site but url.Parse wasn't surfacing the host. anchoring the regex to http(s) only is what keeps `/go:` and `mailto:` from false-tripping, and the userinfo `@` case is a good catch too. in it goes 🖤

PR #173, opened 2026-06-20
fix(scan): key joomla detection on the generator tag

TBX3D:
`detectJoomla` matched any body containing the lowercase word `joomla`, so pages that merely
mention it (joomla hosting marketing like hostinger.com) were misreported. the canonical
`Joomla!` generator tag is capitalised, so the lowercase match also missed sites whose only
tell is that tag. match the `Joomla!` generator meta tag and the joomla asset paths
(`/media/vendor/joomla`, `/media/system/js/core.js`) instead; keying on the meta attribute
rather than the plain tagline also rejects prose that quotes it.

### DIFF
--- a/internal/scan/cms.go
+++ b/internal/scan/cms.go
@@ -87,7 +87,7 @@ func CMS(url string, timeout time.Duration, logdir string) (*CMSResult, error) {
 	}
 
 	// Joomla
-	if strings.Contains(bodyString, "joomla") || strings.Contains(bodyString, "/media/system/js/core.js") {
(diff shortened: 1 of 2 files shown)


vmfunc:
yep, the bare lowercase "joomla" check was matching hosting/marketing pages and somehow still missing the actual capital-J generator tag. keying on the meta generator plus the vendor/core.js paths is the right call, and nice touch 404ing non-root in the test server so the wp probe can't grab the host first. in it goes

PR #174, opened 2026-06-20
fix(scan): drop "query" from generic db admin-panel match

TBX3D:
`isAdminPanel`'s default branch (the generic `/db/`, `/sql/`, `/mongodb/` paths) matched the
keyword `query`, a substring of `jQuery` and `querySelector`, so any site whose catch-all
returns 200/403/401 at one of those paths and runs any javascript was reported as a database
admin panel at high severity. drop `query`; the remaining keywords (`database`, `sql`, `mysql`,
`postgresql`, `mongodb`) only match db-topical pages, so real generic interfaces are still
detected via a sibling keyword.

### DIFF
--- a/internal/scan/sql.go
+++ b/internal/scan/sql.go
@@ -280,10 +280,10 @@ func isAdminPanel(body string, panelType string) bool {
 	case "phpRedisAdmin":
 		return strings.Contains(bodyLower, "phpredisadmin")
 	default:
-		// for generic database interfaces, check for common keywords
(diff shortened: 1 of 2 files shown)


vmfunc:
yep, "query" was matching jquery/queryselector on basically every js page, so every generic /sql/ /db/ soft-404 was firing. clean drop, recall's still covered by the sibling keywords and the e2e jquery catch-all test nails the exact failure. holding on the functional test going green, then in it goes.

PR #175, opened 2026-06-20
fix(scan): flag openapi operations reachable without authentication

TBX3D:
`specToResult` computed an operation's auth status with `len(op.Security) == 0`, which a value
slice cannot tell apart from an absent block: an operation that overrides the global
requirement with an empty array (`security: []`) or an empty requirement object
(`security: [{}]`) decoded the same as one that simply inherits the global default, so those
deliberately-public endpoints were reported as authenticated. decode the operation security
into a pointer to keep absent distinct from explicit-empty, and treat both empty forms as
anonymous-reachable.

### DIFF
--- a/internal/scan/openapi.go
+++ b/internal/scan/openapi.go
@@ -92,10 +92,10 @@ type openapiInfo struct {
 	Version string `json:"version" yaml:"version"`
 }
 
-// rawOps captures just the per-operation security block so we can tell whether
-// an operation requires auth. the rest of the operation object is irrelevant.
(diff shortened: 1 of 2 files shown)


vmfunc:
oh nice, this is the right call. `security: []` and `security: [{}]` opting out of a global requirement were silently passing as authed, and the pointer to tell absent-vs-empty apart is exactly how you fix that. love that the yaml test pins the non-nil empty behavior on the decode path. parking it til ci goes green then it's in.

PR #176, opened 2026-06-20
fix(scan): accept quoted hsts max-age values

TBX3D:
rfc 6797 permits a quoted-string directive value, so a quoted `max-age="31536000"` failed
`strconv.Atoi` and got flagged as too short. strip the quotes before parsing.


vmfunc:
oh nice, the quoted max-age was genuinely mis-grading a perfectly strong hsts as weak. trim before atoi is the right call, and the max-age="0" still-flagged case is the test i'd have asked for. in it goes 🖤

PR #177, opened 2026-06-20
fix(scan): add opt-in soft-404 baseline to db panel scan

TBX3D:
the admin-panel probe trusted any 200/403/401 at a generic db path, so a catch-all that answers every path was reported as a high-severity database admin panel whenever its page merely mentioned a db keyword.

under -ac (the same flag dirlist uses), calibrate the wildcard response shape first and drop probes whose status/size/words match it before isAdminPanel runs. off by default, so the scan keeps reporting every hit unless asked. matching is exact-shape, so a catch-all that reflects the request path is not suppressed.


TBX3D:
superseded by #180 (carries this PR in full). closing.

PR #178, opened 2026-06-20
fix(scan): tolerate reflected-path catch-alls in soft-404 calibration

TBX3D:
the -ac soft-404 baseline matched responses on an exact status/size/words tuple, and the calibration probes were all the same length. a catch-all that echoes the request path answers every path with a different size, so the baseline matched none of the real probe paths and the wildcard leaked.

grow the probe lengths so a reflecting catch-all yields a different size per probe, and when probes share a status and word count but differ in size, record a word-count-tolerant baseline instead of an exact one. stable catch-alls return one shape and keep exact matching, so non-reflecting targets are unchanged. verified live: vercel (200) and a stackoverflow waf (403) reflect the path and are now suppressed under -ac.


TBX3D:
superseded by #180 (carries this PR in full). closing.

PR #179, opened 2026-06-21
fix(scan): don't count a reflected payload as lfi evidence

TBX3D:
the `data://` wrapper payload contains the base64 `PD9waHA` that the base64-php evidence
pattern matches, so a server reflecting the request path was flagged as a critical inclusion.
skip the self-reflected match.

### DIFF
--- a/internal/scan/lfi.go
+++ b/internal/scan/lfi.go
@@ -19,6 +19,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"strings"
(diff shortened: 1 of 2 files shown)


vmfunc:
yeah, reflecting the payload back in an include() warning was tripping the PD9waHA marker, classic false positive. the FindString-not-in-payload guard is the right call and the convert.base64-encode path is untouched so genuine disclosure still fires. in it goes 🖤

PR #180, opened 2026-06-21
fix(scan): apply reflected-path tolerance to db panel calibration

TBX3D:
with `-ac`, the db-panel (sql admin-panel) scan now calibrates a soft-404 baseline the way
dirlist does, so a host that answers 200/403/401 for every path no longer leaks every generic
db path as a panel; real panels keep a distinct word count and still surface. three commits
build it up: dirlist's soft-404 calibration first learns to tolerate catch-alls that echo the
request path (their body size varies per probe, so the baseline compares status and word count
only), then the db-panel scan gains the same opt-in baseline, then its calibration is pointed
at dirlist's shared reflection-tolerant `baselinesFromProbes`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ sif has a modular architecture. modules are defined in yaml and can be extended 
 | `-fs` | dirlist: filter out responses of these body sizes (comma list) |
 | `-fw` | dirlist: filter out responses with these word counts (comma list) |
 | `-fr` | dirlist: filter out responses whose body matches this regex |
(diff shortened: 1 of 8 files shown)


vmfunc:
solid, the reflected-path tolerance is the right gap to close. dirlist_test.go conflicts now, rebase onto main.

vmfunc:
rebased; resolved the dirlist_test keep-both and updated the admin-panel test to the new SQL signature (calibrate defaults false). reflected-path calibration's a solid gap to close. in it goes

PR #181, opened 2026-06-21
fix(scan): scope passive ct-feed subdomains to the target

TBX3D:
shared/multi-SAN certificates list unrelated names, so collecting every crt.sh/certspotter
entry reported off-scope hosts as subdomains of the target. keep only the target or a subdomain
of it.


vmfunc:
yep, multi-SAN certs were dragging in whole unrelated domains, this scopes it cleanly. nice that the notexample.com case nails the suffix-match trap and you caught the trailing-dot fqdn thing too. in it goes

PR #182, opened 2026-06-21
refactor(scan): extract favicon hashing into internal/fingerprint

TBX3D:
splits the shodan favicon hash out of `internal/scan` into a leaf package so the scan check and
the module engine can share one implementation instead of each carrying a copy. pure move, no
behavior change: scan's favicon tests still pass and the golden test moves with the code.

build/vet/lint clean, `go test ./internal/scan/ ./internal/fingerprint/` green.

follow-up that uses this package: #183 (favicon module matcher).


vmfunc:
clean extraction, the murmur3-over-chunked-base64 signedness logic carried over untouched and the golden pins moved with it. scan still owns goldenFaviconBytes for its httptest cases so nothing's orphaned. in it goes once ci's green

PR #183, opened 2026-06-21
feat(modules): add favicon hash matcher

TBX3D:
adds a favicon matcher to the yaml module engine: it matches the shodan-style mmh3 hash of the
response body, accepts the hash signed or unsigned, validates it at load so a bad value fails
loudly, and reports the hash (not raw icon bytes) as evidence for a favicon-only finding.
stacked on #182 (the first commit here is that fingerprint extraction); once it lands i'll
rebase this onto main so only the feat commit remains.

build/vet/lint clean, `go test ./internal/modules/` green (matcher, signed/unsigned normalize,
load validation, httptest end-to-end).

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -223,6 +223,30 @@ matchers:
       - 1337
 ```
 
+### favicon matcher
+
+match the shodan-style mmh3 hash of the response body. point the module at a
+favicon and list the hashes of the tech you want to fingerprint.
+
+```yaml
+http:
+  paths:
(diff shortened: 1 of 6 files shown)


vmfunc:
the favicon hash matcher's good. it conflicts across module.go/yaml.go with the engine PRs that landed, rebase onto main and it'll replay.

vmfunc:
rebased; the favicon hash matcher merged in alongside the engine changes that landed. in it goes

PR #184, opened 2026-06-21
feat(modules): add favicon fingerprint demo module

TBX3D:
stacked on #183 (and #182); until those land this pr's diff includes their commits, and i'll
rebase onto main once they merge so it reduces to the single demo commit.
`modules/info/favicon-gitlab.yaml` fingerprints gitlab by its default `/favicon.ico` hash using
the favicon matcher from #183, and `internal/scan/favicon_demo_sync_test.go` ties the demo hash
to the canonical `faviconHashes` map so the yaml cannot drift from the scanner's source of
truth. matches `/favicon.ico` default installs; instances that serve their icon from an
asset-pipeline path via `<link rel=icon>` are out of scope.

build/vet/lint clean, `go test ./internal/modules/ ./internal/scan/` green.

### DIFF
--- a/internal/scan/favicon_demo_sync_test.go
+++ b/internal/scan/favicon_demo_sync_test.go
@@ -1,0 +1,68 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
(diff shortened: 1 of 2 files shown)


vmfunc:
depends on #183, same rebase. once #183 is replayed onto main, rebase this on top of it.

vmfunc:
rebased onto main on top of #183. favicon fingerprint demo, in 🖤

PR #185, opened 2026-06-21
feat(modules): add env file exposure module

TBX3D:
detects exposed `.env` files across the common variant paths (`.env`, `.env.local`,
`.env.production`, `.env.backup`, ...). fires on a 200 whose body carries a secret-bearing
`KEY=` line (`DB_PASSWORD=`, `APP_KEY=`, `AWS_SECRET_ACCESS_KEY=`, `JWT_SECRET=`) and reports
which key leaked. a negative matcher on `<!DOCTYPE`/`<html` keeps an html page that merely
mentions a key from matching, the usual false positive here.

build/vet/lint clean, `go test ./internal/modules/` green (real-leak, html-mention, 404).

### DIFF
--- a/internal/modules/env_exposure_test.go
+++ b/internal/modules/env_exposure_test.go
@@ -1,0 +1,84 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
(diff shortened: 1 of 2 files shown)


vmfunc (changes requested):
env_exposure_test.go opens straight on `package modules_test` with no banner. check-headers skips _test.go so it didn't catch it, but every other .go in the tree carries the BSD banner block (grab it off matchers_test.go). prepend it and this is good to go, rest of the test reads clean.

TBX3D:

vmfunc:
banner's in, thanks. env detection reads clean, merged 🖤

PR #186, opened 2026-06-21
feat(modules): add adminer and phpmyadmin database panel modules

TBX3D:
two info-level modules that flag exposed database admin panels. adminer keys on the login form's
`auth[...]` field names and the `adminer.org` footer link, and extracts the version from the
footer. phpmyadmin keys on the `pma_username`/`pma_password` fields, the `pmahomme` theme path
and the `phpMyAdmin` cookie, and extracts `PMA_VERSION`. both gate on a 200 so a bare brand
mention on an unrelated page does not match.

build/vet/lint clean, `go test ./internal/modules/` green (real-hit and unrelated-page cases).

### DIFF
--- a/internal/modules/db_panels_test.go
+++ b/internal/modules/db_panels_test.go
@@ -1,0 +1,88 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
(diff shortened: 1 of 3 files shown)


vmfunc:
clean batch. both modules gate on status + a word matcher so they won't fire on a bare 200, and using part: all on the pma extractor to catch the phpMyAdmin= cookie + pmahomme is the right call. tests actually exercise the cookie path too. in it goes 🖤

PR #187, opened 2026-06-21
feat(modules): add grafana, kibana and jenkins login panel modules

TBX3D:
three info-level modules for exposed ops and ci login panels. grafana keys on
`window.grafanaBootData` and the `grafana-app` shell, and pulls the version from `buildInfo`.
kibana keys on the `kbn-version`/`kbn-name` response headers and the injected-metadata markers,
and pulls the version from the header. jenkins keys on the `X-Jenkins` header (present even when
`/` returns 403) and the login form, and pulls the version from that header. each module's
detection and version come from the same response, so the extractor only runs on a real hit.

build/vet/lint clean, `go test ./internal/modules/` green (including the kibana and jenkins
response-header paths).

### DIFF
--- a/internal/modules/login_panels_test.go
+++ b/internal/modules/login_panels_test.go
@@ -1,0 +1,106 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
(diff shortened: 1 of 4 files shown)


vmfunc (changes requested):
login_panels_test.go is missing the banner, it starts cold at `package modules_test`. CI won't flag it since check-headers skips _test.go, but house rule is every .go file gets the block. copy it off executor_test.go and i'll land it. rest of the test's fine.

TBX3D:

vmfunc:
fixed, ty. grafana/kibana/jenkins panel detection is tight. in it goes

PR #188, opened 2026-06-21
feat(frameworks): add htmx detector

TBX3D:
adds an htmx detector to the frontend framework set. it scores the declarative `hx-*` attributes
(`hx-get`, `hx-post`, `hx-swap`, `hx-target`, `hx-boost`) plus the `htmx.org` library reference;
the weights are sized so two htmx signals clear the detection threshold but a lone stray `hx-`
substring does not, matching how the sibling detectors avoid single-signal false positives.
version comes from the cdn or package reference (`htmx.org@x.y.z`). nuxt already has a detector
in `meta.go`, so this adds htmx only.

build/vet/lint clean, `go test ./internal/scan/frameworks/...` green (detection on a
representative htmx page, version extraction).

### DIFF
--- a/internal/scan/frameworks/detect_test.go
+++ b/internal/scan/frameworks/detect_test.go
@@ -518,3 +518,55 @@ func TestDetectorRegistry(t *testing.T) {
 		}
 	}
 }
+
+func TestExtractVersion_Htmx(t *testing.T) {
+	tests := []struct {
+		body     string
+		expected string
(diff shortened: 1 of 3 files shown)


vmfunc:
clean little detector, the hx-* signature set plus the cdn @version regex is exactly the right shape and it mirrors the meteor one perfectly. in it goes once ci's green

PR #189, opened 2026-06-21
feat(modules): support or logic via matchers-condition

TBX3D:
`checkMatchers` combined matchers with AND only. add a `matchers-condition` field accepting
`and` (default) or `or` so an http module can fire when any matcher matches instead of all;
invalid values are rejected at load.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -242,6 +242,25 @@ matchers:
 
 this matches responses with status 200 AND containing "ref: refs/".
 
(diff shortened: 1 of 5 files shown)


vmfunc:
the or-logic via matchers-condition is a good add. yaml.go conflicts now, rebase onto main.

vmfunc:
rebased onto main, kept both the pitchfork attack validation and your matchers-condition validation. or-logic is a good add. in 🖤

PR #190, opened 2026-06-21
feat(modules): add portainer, traefik, keycloak and rabbitmq panel modules

TBX3D:
four info-level modules for exposed ops admin panels and dashboards. portainer and traefik
detect via their unauthenticated json apis (`/api/status` and `/api/version`) and extract the
reported version. keycloak detects via its public `/realms/master` endpoint, and rabbitmq via
the management ui title. matchers gate on a 200 and require the product-specific keys together
(portainer needs `Edition` + `Version` + `InstanceID`, for one), so a generic json body with a
stray `Version` field does not match.

build/vet/lint clean, `go test ./internal/modules/` green (each module end to end, real-hit and
near-miss).

### DIFF
--- a/internal/modules/ops_panels_test.go
+++ b/internal/modules/ops_panels_test.go
@@ -1,0 +1,112 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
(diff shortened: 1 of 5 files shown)


vmfunc:
clean batch. anding the status matcher with the multi-field body markers is the right move, the "real body behind a 404" negative test is exactly the gap that bites these panel fingerprints. rabbitmq one is the loosest since it hits bare baseurl but the logo/title words keep it honest. green ci and it's in 🖤

PR #191, opened 2026-06-21
feat(modules): add json extractor

TBX3D:
`runExtractors` only handled regex. add a json extractor that pulls values from a json body by
gjson path, storing the first path that exists under the extractor name. gjson was already an
indirect dependency.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -271,6 +271,21 @@ extractors:
     part: header
 ```
 
+### json extractor
+
+extract values from a json body by gjson path (github.com/tidwall/gjson); the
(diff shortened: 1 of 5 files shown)


vmfunc:
json extractor's clean, gjson was already vendored. it conflicts with the kv extractor in executor.go now: #140 moved getPart inside each case, so on replay the json case needs its own getPart. rebase onto main and it's in 🖤

vmfunc:
rebased onto main. #140's kv extractor moved getPart into each case, so i gave the json case its own getPart on replay. gjson was already vendored, basically free. in 🖤

PR #192, opened 2026-06-21
feat(modules): add spring actuator, phpinfo and apache status exposure modules

TBX3D:
`modules/recon/actuator-exposure.yaml` flags publicly exposed spring boot actuator endpoints
(env, health, configprops and the legacy unprefixed paths) on the structural json markers
rather than the path alone, so a bare 200 is not a hit. `modules/recon/phpinfo-exposure.yaml`
requires the generated `<title>phpinfo()` on top of the `PHP Version` string, so a page that
only mentions the version in prose is not a false positive.
`modules/recon/apache-status-exposure.yaml` covers mod_status and mod_info, including the
`?auto` plain-text variant, and extracts the server version.

`ExecuteHTTPModule`, hit and near-miss; each matcher/extractor proven to bite).

### DIFF
--- a/internal/modules/diagnostic_exposure_test.go
+++ b/internal/modules/diagnostic_exposure_test.go
@@ -1,0 +1,107 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+// runDiagModule runs a shipped module end to end against a server that returns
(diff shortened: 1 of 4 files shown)


TBX3D:
closing as a duplicate: spring actuator, phpinfo and apache status are already covered by the focused single-service prs #144, #145 and #150. the only coverage here not already in those, apache `mod_info` (`/server-info`) and the `?auto` status variant plus the spring boot 1.x top-level `/env` and `/health` paths, has been folded into #150 and #144, so nothing is lost.

PR #193, opened 2026-06-21
feat(modules): add pitchfork attack mode

TBX3D:
the `attack` field was parsed but never read, so every module ran the clusterbomb
cross-product. wire it up: `pitchfork` pairs `path[i]` with `payload[i]` and stops at the
shorter list, `clusterbomb` stays the default, and an unknown attack value is rejected at
parse time instead of silently ignored. `sniper` is deliberately left out: a faithful sniper
needs multiple payload positions with base values, which the single flat `payloads` list and
one `{{payload}}` token can't model, so it would collapse into clusterbomb for >1 path; it
can follow once named payload sets exist.

build/vet/lint clean, `go test ./internal/modules/` green (generator even/uneven, validator,
parse-time rejection, end-to-end pitchfork).

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -115,6 +115,18 @@ http:
 
 each payload creates a separate request for each path.
 
+#### attack
+
+how paths and payloads combine into requests.
+
+```yaml
+http:
+  attack: pitchfork
+```
+
+- `clusterbomb` (default) - every path is tried with every payload
+- `pitchfork` - path and payload are paired by index, stopping at the shorter list
+
 #### headers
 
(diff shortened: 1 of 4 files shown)


vmfunc:
pitchfork pairing by index with the min-length stop is exactly right, and the end-to-end test proving it fires /a?x=1 and /b?x=2 but not the cross product is the part i'd have asked for. validating attack at parse time so a typo'd mode fails loud instead of silently falling back to clusterbomb is a nice touch. holding for ci to go green since it's in the request-gen core, then in it goes.

PR #62, opened 2026-01-07
feat(modules): builtin shodan scan as module

0x4bs3nt:
First part of completing the #52 issue by making requested complex Go scans available as modules.

## Summary
 - Allows complex Go scan; shodan, to participate in the sif module system

## Features
 - Made Go scan for shodan available as a module
 - Added to module registry
 - Will show up in the `-lm/-list-modules`

## Test plan

- [x] `gofmt` applied
- [x] all unit tests passing
- [x] `go vet` passes
- [x] manually tested shodan works the same as a module and as a legacy Go scan

Issue #50, opened 2026-01-03
remove references to lunchcat domain


vmfunc:
[makefile](https://github.com/vmfunc/sif/blob/8fb797dd3a8ab936eeb233f7f118259758e0317c/Makefile) still holds reference to the lunchcat domain - it would be good to go through all the files and check for old references to lunchcat and especially discontinued domains that are now inactive.

copyright to lunchcat will be kept for the sake of continuity but old domains or erroneous information should be removed

PR #194, opened 2026-06-21
chore: migrate module path to github.com/vmfunc/sif

TBX3D:
migrate the go module path from `github.com/dropalldatabases/sif` to `github.com/vmfunc/sif`
for package-manager and semver compliance. refs #57; redo of #59, which was closed unmerged
behind #62/#63/#64 (all merged now). atomic by necessity: `go.mod` and every import move in a
single commit so each commit still builds; this exceeds the 5-file guideline because a go
module rename cannot be split without breaking intermediate builds. the sif-runtime data urls
are intentionally left on `dropalldatabases/sif-runtime`: `vmfunc/sif-runtime` does not exist
(404) and the current path still serves the data (it redirects to `lunchcat/sif-runtime`), so
moving them would break runtime fetches for dnslist, dork, dirlist, git and ports. out of
scope: lunchcat domain refs (#50), the v1.0.0 tag, CHANGELOG and discord-link standardization.

build/vet/lint/gofmt clean, `go test` green; binary builds and runs.


vmfunc:
this is the right migration and it should land, but it needs re-running across the full tree, ~90 files reference the old path now (not the 66 when you opened it), and it has to go in LAST once the other open PRs are rebased or it'll conflict with every one of them. rebase onto main, re-run the rename, and ping me to land it 🖤

vmfunc:
landed the migration over the full tree, re-ran the rename so it covers everything that merged after you opened it (102 files now). go module path is github.com/vmfunc/sif. thanks for driving this 🖤

PR #195, opened 2026-06-21
feat(modules): add aws, npmrc and docker credential file exposure modules

TBX3D:
`modules/recon/aws-credentials-exposure.yaml` flags exposed `.aws/credentials`, `.s3cfg` and
`.boto` files on the access and secret key markers, and extracts the `AKIA`/`ASIA` access key
id. `modules/recon/npmrc-exposure.yaml` flags a `.npmrc` only when it carries an auth token or
password, not a bare registry config, and extracts the registry the token belongs to.
`modules/recon/docker-config-exposure.yaml` flags `.docker/config.json` and the legacy
`.dockercfg` on the base64 `auth` field, and extracts the registry host. each module ands a
negative matcher on the usual html markers so a 200 page that merely names a key is not a hit,
the same guard the env exposure module uses.

build/vet/lint clean, `go test ./internal/modules/` green (the three modules end to end via `ExecuteHTTPModule`;
near-misses pinned: html naming a key, plain 200, 404, a jwt-shaped docker auth value).

### DIFF
--- a/internal/modules/credential_exposure_test.go
+++ b/internal/modules/credential_exposure_test.go
@@ -1,0 +1,113 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+func runCredModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


vmfunc:
clean batch. the negative html matchers are the right call, that's what keeps the docs-page-mentioning-aws_secret_access_key noise out, and the docker regex correctly skips a jwt in an auth field since the dots break the base64 run. tests cover the soft-404 and plain-200 gaps too. in it goes 🖤

PR #196, opened 2026-06-21
feat(modules): add ignition, profiler and heapdump exposure modules

TBX3D:
`modules/recon/laravel-ignition-exposure.yaml` probes the live `/_ignition/health-check`
endpoint and extracts `can_execute_commands`, the flag that marks the CVE-2021-3129 rce
surface; an active probe complementary to the version-based ignition entry in the framework cve
map. `modules/recon/symfony-profiler-exposure.yaml` flags an exposed web profiler on its
structural markers and extracts a request token to pivot to a captured request.
`modules/recon/spring-heapdump-exposure.yaml` flags an exposed actuator heap dump on the hprof
magic anchored at the start of the body (which a json-marker module cannot see, the dump being
binary) and extracts the hprof version; the anchor keeps a page that merely quotes the magic
from matching.

near-misses: prose mention, hprof magic off-start, plain 200, 404; plus ignition with
command-exec disabled still flags and reports the false flag).

### DIFF
--- a/internal/modules/debug_exposure_test.go
+++ b/internal/modules/debug_exposure_test.go
@@ -1,0 +1,121 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+func runDebugModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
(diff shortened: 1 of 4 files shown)


vmfunc:
these are tight. nice touch anchoring the hprof magic with ^JAVA PROFILE so a prose mention of the header doesn't false-positive, and the test that proves it. once functional test + yamllint go green this is in.

PR #197, opened 2026-06-21
fix(scan): crack weak hmac secrets on hs384 and hs512 tokens

TBX3D:
the jwt weak-secret cracker always hashed with sha-256 while `crackHMAC` was called for the
whole hmac family (`isHMACAlg` accepts HS256/HS384/HS512). on an HS384/HS512 token it compared
a 32-byte digest against a 48/64-byte signature, so `hmac.Equal` could never match and
weak-secret detection silently never fired on those tokens. `crackHMAC` now selects the digest
by alg via a new `hmacHash` helper (HS256 -> sha256, HS384 -> sha512.New384, HS512 ->
sha512.New); an unknown or non-hmac alg cracks nothing. `hmacHash` is intentionally stricter
than `isHMACAlg`, which still flags any HS* alg for the confusion-surface finding.

build/vet/lint clean, `go test ./internal/scan/` green (HS384/HS512 weak-secret cracks proven
red-before/green-after, strong-secret negatives, an `hmacHash` digest-width table, a malformed
no-panic case).

### DIFF
--- a/internal/scan/jwt.go
+++ b/internal/scan/jwt.go
@@ -16,9 +16,11 @@ import (
 	"context"
 	"crypto/hmac"
 	"crypto/sha256"
+	"crypto/sha512"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"hash"
 	"io"
 	"net/http"
 	"regexp"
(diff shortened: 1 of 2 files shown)


vmfunc:
nice, the hs256-only crack path was leaving hs384/hs512 weak secrets on the table. good detail keeping hmacHash stricter than isHMACAlg so the confusion finding still fires but cracking only runs on a real digest width. verified the test vectors against "secret" myself, in it goes 🖤

PR #198, opened 2026-06-21
feat(modules): add editor sftp deploy config exposure modules

TBX3D:
`modules/recon/vscode-sftp-exposure.yaml` flags an exposed vscode-sftp config on its tool keys,
`remotePath` and `uploadOnSave`, then extracts the deploy host; the tool keys keep an unrelated
json config that merely carries host and credential fields from matching.
`modules/recon/sublime-sftp-exposure.yaml` flags an exposed Sublime SFTP config on its
snake-case keys, `upload_on_save` and `sync_down_on_open`, and extracts the deploy host.
`modules/recon/ftpconfig-exposure.yaml` flags an exposed remote-ftp config on its
connection-timeout keys, `connTimeout` and `pasvTimeout`, and extracts the deploy host. each
module requires a credential field alongside the tool key and rejects an html body, so a login
page on the same path is not a leak and an unrelated json config is not a high-severity
credential finding.

near-misses pinned: html login page with the same keys, plain json without the tool keys, a
tool config with host but no credential, a 404; plus a key-auth config with no password still
caught as a leak).

### DIFF
--- a/internal/modules/deploy_config_exposure_test.go
+++ b/internal/modules/deploy_config_exposure_test.go
@@ -1,0 +1,134 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+func runDeployModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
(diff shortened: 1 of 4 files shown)


vmfunc:
these are tight. the negative html matcher killing the login-page false positive is the part that matters and the test for it is exactly right. key-auth vscode configs (privateKeyPath, no password) still flagging is a nice touch too. in it goes 🖤

PR #199, opened 2026-06-21
feat(modules): add rails config secret exposure modules

TBX3D:
`modules/recon/rails-database-yml-exposure.yaml` flags an exposed `config/database.yml` on the
adapter key paired with a credential key, then extracts the database name; requiring a
credential key keeps a credential-free sqlite config from being reported as a high-severity
leak. `modules/recon/rails-secrets-yml-exposure.yaml` flags an exposed `config/secrets.yml` on
the `secret_key_base` key and extracts the secret an attacker needs to forge rails sessions.
`modules/recon/rails-master-key-exposure.yaml` flags an exposed master key, the 32-hex value
that decrypts the encrypted credentials store; the matcher anchors the hex to the whole body so
a longer digest (a sha256 served at the same path) does not match, and the same probe covers
`config/credentials`.

near-misses pinned: credential-free sqlite config, a longer hex digest, hex off body-start,
html naming the markers, a config without the markers, a 404).

### DIFF
--- a/internal/modules/rails_secret_exposure_test.go
+++ b/internal/modules/rails_secret_exposure_test.go
@@ -1,0 +1,131 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+func runRailsModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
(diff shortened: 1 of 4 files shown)


vmfunc:
these are tight. the negative-true html guard on the word matchers kills the soft-404 false positives, and the whole-string-anchored ^[a-f0-9]{32}$ on the master key means a 64-char digest won't trip it. tests cover the sqlite-no-creds case too, which is the easy one to miss. in it goes 🖤

PR #200, opened 2026-06-21
fix(scan): don't follow redirects when probing cors

TBX3D:
the cors probe followed up to three redirects, then read the cors headers off the final
response. its own comment said the cap existed so the headers came from the host we asked
about, so the code contradicted its intent: a target that only redirects to a permissive third
party had that party's misconfiguration reported against it. stop following redirects, matching
the open-redirect scanner's no-follow behavior. the redirect cap was a const shared with the
xss probe, so xss now owns its own `xssMaxRedirects` with its follow behavior unchanged.

build/vet/lint clean, `go test ./internal/scan/` green
(`TestCORS_JudgesRequestedHostNotRedirectTarget`: a host that only bounces to a reflecting
third party yields no finding, an atomic hit-counter proves the probe never leaves the
requested host; fails on the old code with six misattributed findings).

### DIFF
--- a/internal/scan/cors.go
+++ b/internal/scan/cors.go
@@ -42,10 +42,6 @@ type CORSFinding struct {
 	Note             string `json:"note"`
 }
 
-// corsMaxRedirects caps the redirect chain so we read the cors headers off the
-// host we actually asked about, not whatever it bounces us to.
-const corsMaxRedirects = 3
-
 // the sentinel attacker origin; if it comes back in Access-Control-Allow-Origin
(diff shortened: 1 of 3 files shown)


vmfunc:
yep, following the bounce meant a redirect to some permissive third party got pinned on the target as if it were the target's cors policy. dropping to ErrUseLastResponse is right, and splitting xssMaxRedirects out so xss keeps its 3-hop follow is the clean way to do it. the destHits atomic counter in the test nails exactly the regression. in it goes 🖤

PR #201, opened 2026-06-21
feat(modules): add terraform, kubeconfig and compose exposure modules

TBX3D:
`modules/recon/terraform-state-exposure.yaml` flags an exposed terraform state file on the
`terraform_version` key paired with a state-structure key, then extracts the version; the
structure key keeps a document that merely mentions `terraform_version` from matching.
`modules/recon/kubeconfig-exposure.yaml` flags an exposed kubeconfig on the `kind: Config`
marker paired with a cluster or credential key, then extracts the cluster api endpoint (an
exec-auth kubeconfig with no embedded key still leaks via the cluster block alone).
`modules/recon/docker-compose-exposure.yaml` flags an exposed compose file on the `services`
key paired with a service-definition key, then extracts the first image reference to surface
the stack and its versions. each module pairs a unique marker with a structure key and rejects
an html body, so a page that only names the marker is not a leak.

`ExecuteHTTPModule`; near-misses pinned: bare `terraform_version`/`kind: Config`/`services`
mentions, html carrying the markers, plain 200, 404).

### DIFF
--- a/internal/modules/infra_config_exposure_test.go
+++ b/internal/modules/infra_config_exposure_test.go
@@ -1,0 +1,151 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
+:   █▀ █ █▀▀   ·   Blazing-fast pentesting suite                                :
+:   ▄█ █ █▀    ·   BSD 3-Clause License                                         :
+:                                                                               :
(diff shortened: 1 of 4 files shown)


vmfunc (changes requested):
infra_config_exposure_test.go opens straight on `package modules_test`, no banner. every test in internal/modules/ has the block (matchers_test.go etc), check-headers just happens to skip _test.go so it slid through green. add it and this is good. file's otherwise clean.

TBX3D:

vmfunc:
banner sorted, merged. the terraform/kubeconfig/compose set is solid

PR #202, opened 2026-06-21
feat(modules): add htpasswd, web.config and htaccess exposure modules

TBX3D:
`modules/recon/htpasswd-exposure.yaml` flags an exposed htpasswd file on a line holding a
recognised password hash (apache md5, bcrypt, sha-crypt or a `{SHA}` digest), then extracts the
user; matching the hash format keeps a plaintext line from being reported.
`modules/recon/webconfig-exposure.yaml` flags an exposed asp.net `web.config` on the
configuration root paired with a dotnet section, then extracts a connection string (the value
carrying the database server and password). `modules/recon/htaccess-exposure.yaml` flags an
exposed htaccess file on its apache directives and extracts the `AuthUserFile` path, which
points at the password file to fetch next.

`ExecuteHTTPModule`; near-misses pinned: a plaintext htpasswd line, a configuration without a
dotnet section, an html page, plain 200, 404).

### DIFF
--- a/internal/modules/webserver_config_exposure_test.go
+++ b/internal/modules/webserver_config_exposure_test.go
@@ -1,0 +1,136 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+func runWebSrvModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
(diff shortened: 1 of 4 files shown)


vmfunc:
these are tight, the htaccess negative matcher on <!doctype/<html to kill soft-404 echoes is the part that makes them usable, and the htpasswd regex actually distinguishing apr1/bcrypt from a plaintext line is the right call. tests cover the fp cases too. in it goes 🖤

PR #203, opened 2026-06-21
fix(scan): don't flag quotes reflected in text as attribute xss

TBX3D:
`classifyXSSContext` defaulted any non-html, non-script reflection to attribute context. a
reflection echoed into element text with the angle brackets escaped but quotes left raw (common
for encoders limited to `< > &`) was then reported as a high-severity attribute injection, even
though a quote in text content has no delimiter to break out of. classify a reflection as
attribute only when the canary actually lands inside an open tag; everything else is inert
element text and yields no finding. the in-tag test is a byte-scan, not a parser, so rare
malformed markup can still mis-bucket.

### DIFF
--- a/internal/scan/xss.go
+++ b/internal/scan/xss.go
@@ -234,8 +234,9 @@ func probeXSS(client *http.Client, parsedURL *url.URL, existing url.Values, para
 
 // classifyXSSContext guesses where the canary was reflected. We look at the
 // markup immediately around the token: a live <canary> tag means html text, a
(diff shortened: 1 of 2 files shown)


vmfunc:
nice, the default-to-attribute bucket was exactly the bug, a quote in escaped element text breaks out of nothing. reflectedInsideTag gating it plus the new text->empty-set path is the right call, and good that you kept a real attribute sink in the tests to prove the true positive still fires. in it goes.

PR #204, opened 2026-06-21
feat(modules): add sql dump, sqlite and redis rdb exposure modules

TBX3D:
`modules/recon/sql-dump-exposure.yaml` flags an exposed sql dump on its `mysqldump` and
`pg_dump` idioms paired against a guard that drops sql shown inside an html page, then extracts
the dumped table name. `modules/recon/sqlite-database-exposure.yaml` flags an exposed sqlite
file on the 16-byte format magic anchored to the start of the body, then extracts a schema
table name; anchoring the magic keeps a page that merely embeds the header from being reported.
`modules/recon/redis-dump-exposure.yaml` flags an exposed redis rdb snapshot on the RDB magic
anchored to the start of the body, then extracts the format version.

`ExecuteHTTPModule`; near-misses pinned: a sql tutorial page, a bare select, prose naming the
sqlite/redis format, a header embedded mid-body, plain 200, 404).

### DIFF
--- a/internal/modules/database_file_exposure_test.go
+++ b/internal/modules/database_file_exposure_test.go
@@ -1,0 +1,166 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+func runDBFileModule(t *testing.T, file string, status int, body string) *modules.Result {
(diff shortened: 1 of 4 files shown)


vmfunc:
oh these are tight. the negative-html guard on the sql one plus the embedded-magic-in-prose negatives are exactly the false-positive gaps most modules skip, and gating sqlite/redis on the magic at offset 0 is the right call. in it goes once ci's green 🖤

PR #205, opened 2026-06-21
feat(modules): add private key, git and pypi secret exposure modules

TBX3D:
`modules/recon/private-key-exposure.yaml` flags an exposed PEM private key on the
`BEGIN PRIVATE KEY` marker (so a public key or prose that merely names a key is left alone),
then extracts the key type. `modules/recon/git-credentials-exposure.yaml` flags an exposed git
credential store on a remote url carrying an inline password, paired with a guard that drops a
url shown inside an html page, then extracts the host the credential reaches.
`modules/recon/pypirc-exposure.yaml` flags an exposed pypirc on an index section paired with a
credential field, then extracts the pypi upload token; requiring the credential keeps a bare
index listing from being reported.

`ExecuteHTTPModule`; near-misses pinned: a public key, prose naming a key, a remote url with no
password, a pypi section with no credential, a credential in an html page, plain 200, 404).

### DIFF
--- a/internal/modules/secret_file_exposure_test.go
+++ b/internal/modules/secret_file_exposure_test.go
@@ -1,0 +1,162 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+func runSecretModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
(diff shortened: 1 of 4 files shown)


vmfunc:
these are tight. the negative html-guard on the git/pypi matchers kills the obvious false-positive where a docs page just shows a user:pass@host clone string, and the false-positive matrix (public-key-not-private, bare remote, 404) is exactly the stuff that usually makes exposure modules noisy. extractors pulling git_host/key_type/pypi_token are a nice touch. in it goes, ty tigah 🖤

PR #206, opened 2026-06-21
feat(modules): add spring, appsettings and wp-config exposure modules

TBX3D:
`modules/recon/spring-application-config-exposure.yaml` flags an exposed spring application
config (properties or yaml) on a datasource marker paired with a credential field, then
extracts the jdbc url; requiring the credential keeps a secret-free config from being reported.
`modules/recon/appsettings-exposure.yaml` flags an exposed asp.net core `appsettings.json` (the
.net core counterpart to `web.config`) on a `ConnectionStrings` section paired with an inline
password, then extracts the connection string. `modules/recon/wp-config-backup-exposure.yaml`
flags an exposed wp-config backup (the leftover `.bak` or swap copy that serves raw php) on the
`DB_PASSWORD` constant paired with another db define, then extracts the database password.

`ExecuteHTTPModule`; near-misses pinned: a config with no credential, a password outside a
connection-strings section, a passwordless connection string, prose naming `DB_PASSWORD`, a
config in an html page, plain 200, 404).

### DIFF
--- a/internal/modules/app_config_exposure_test.go
+++ b/internal/modules/app_config_exposure_test.go
@@ -1,0 +1,173 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+func runAppCfgModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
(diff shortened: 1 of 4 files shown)


vmfunc:
these are tight. the negative html guards on spring and wp-config are exactly right, a rendered page with DB_PASSWORD in it shouldn't fire, and the appsettings ConnectionStrings + password gating kills the passwordless-string false positives. in it goes once ci's green 🖤

PR #207, opened 2026-06-21
feat(modules): add vault, consul and etcd api exposure modules

TBX3D:
`modules/recon/vault-api-exposure.yaml` flags an internet-reachable hashicorp vault through its
unauthenticated seal-status, keyed on the `sealed` flag paired with a vault-only status field,
then extracts the version for cve matching. `modules/recon/consul-api-exposure.yaml` flags a
consul http api that answers without an acl token, keyed on the `Datacenter` field paired with
an agent or node marker, then extracts the datacenter. `modules/recon/etcd-api-exposure.yaml`
flags an exposed etcd through its version endpoint, keyed on the `etcdserver` and `etcdcluster`
fields a generic version response does not carry, then extracts the server version.

`ExecuteHTTPModule`; near-misses pinned: a sealed flag with no vault field, a datacenter field
alone, a version response from another service, a partial etcd reply, plain 200, 404).

### DIFF
--- a/internal/modules/orchestration_api_exposure_test.go
+++ b/internal/modules/orchestration_api_exposure_test.go
@@ -1,0 +1,132 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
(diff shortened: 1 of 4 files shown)


vmfunc:
these are tight. the etcdserver-AND-etcdcluster gating is the right call, a lone version field would've been noisy as hell, and the negative tests prove it. holding only til ci goes green, then in it goes. ty tigah 🖤

PR #208, opened 2026-06-21
feat(modules): add netrc, pgpass and my.cnf exposure modules

TBX3D:
`modules/recon/netrc-exposure.yaml` flags an exposed `.netrc` on a
`machine ... login ... password ...` line, then extracts the machine.
`modules/recon/pgpass-exposure.yaml` flags an exposed `.pgpass` on its
`host:port:db:user:password` line format, then extracts the host.
`modules/recon/mysql-client-config-exposure.yaml` flags an exposed `.my.cnf` on a
`[client]`/`[mysql]` section paired with a `password=` line, then extracts the user. each gates
on a 200 and rejects an html body, so a 200 page that merely names the file is not a leak.

### DIFF
--- a/internal/modules/dotfile_credential_exposure_test.go
+++ b/internal/modules/dotfile_credential_exposure_test.go
@@ -1,0 +1,164 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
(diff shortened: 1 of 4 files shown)


vmfunc:
these are tight, the negative html guard doing the work on the tutorial-page false positives is exactly right and the pgpass numeric-port regex is a nice touch. holding only til the functional test goes green, then in it goes 🖤

PR #209, opened 2026-06-21
feat(modules): add maven, gradle and nuget credential exposure modules

TBX3D:
`modules/recon/maven-settings-exposure.yaml` flags an exposed maven `settings.xml`
(`.m2/settings.xml` or root) on a `<servers>` section paired with a `<password>`, then extracts
the `<username>`. `modules/recon/gradle-properties-exposure.yaml` flags an exposed
`gradle.properties` on a `password`/`secret`/`token` property line, then extracts that property
name. `modules/recon/nuget-config-exposure.yaml` flags an exposed `nuget.config` on a
`packageSourceCredentials` section paired with a `ClearTextPassword` or `Password`, then
extracts the username. each gates on a 200 and rejects an html body, so a page that merely names
the file is not a leak.

### DIFF
--- a/internal/modules/buildtool_credential_exposure_test.go
+++ b/internal/modules/buildtool_credential_exposure_test.go
@@ -1,0 +1,198 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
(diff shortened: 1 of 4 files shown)


vmfunc (changes requested):
buildtool_credential_exposure_test.go goes straight to `package modules_test` with no banner. check-headers skips _test.go so it's green, but every .go in the repo carries the block, copy it off matchers_test.go in the same package. rest is clean, terse subtests, no other notes.

TBX3D:

vmfunc:
in. nice work on the maven/gradle/nuget credential modules

PR #210, opened 2026-06-21
feat(modules): add svn, mercurial and bazaar exposure modules

TBX3D:
`modules/recon/svn-exposure.yaml` flags an exposed `.svn` working copy through the `wc.db`
sqlite header anchored at the first byte paired with a `WCROOT` or `PRISTINE` table name, so a
generic sqlite database (or an html page embedding the magic) is not reported, then extracts the
repository url. `modules/recon/mercurial-exposure.yaml` flags an exposed `.hg` repository on the
revlog format requirements the `requires` file lists (`revlogv1`, `dotencode`, `fncache`), so
prose naming mercurial or an html tutorial showing the format is not reported, then extracts the
requirement. `modules/recon/bazaar-exposure.yaml` flags an exposed `.bzr` repository on its
`Bazaar` `meta directory` branch-format signature, so a marketplace page that merely names a
bazaar is not reported, then extracts the format.

### DIFF
--- a/internal/modules/vcs_metadata_exposure_test.go
+++ b/internal/modules/vcs_metadata_exposure_test.go
@@ -1,0 +1,155 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+func runVCSModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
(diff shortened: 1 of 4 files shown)


vmfunc:
clean batch. the `^SQLite format 3\x00` anchor plus the WCROOT/PRISTINE gate is exactly right for not eating random sqlite dumps, and the svn url extractor stopping at \x00 keeps it from running past the null-delimited field. the negative doctype/html guards on hg and bzr are a nice touch, those branch-format/requires files get mirrored into tutorials constantly. merged 🖤

PR #211, opened 2026-06-21
feat(modules): add joomla, drupal and magento config exposure modules

TBX3D:
`modules/recon/joomla-config-exposure.yaml` flags an exposed joomla `configuration.php` backup
on a `public $password` property paired with the `JConfig` class marker, so a generic php class
is not reported, then extracts the database password. `modules/recon/drupal-config-exposure.yaml`
flags an exposed drupal `settings.php` backup on the `$databases` array paired with a literal
`'password' => '...'` value, so an array that resolves the password from the environment is not
reported, then extracts the password. `modules/recon/magento-config-exposure.yaml` flags an
exposed magento `app/etc/env.php` backup on a `crypt` or `MAGE_MODE` marker paired with a literal
key or password value, so a generic return array or a cloud placeholder is not reported, then
extracts the crypt key.

### DIFF
--- a/internal/modules/cms_config_exposure_test.go
+++ b/internal/modules/cms_config_exposure_test.go
@@ -1,0 +1,205 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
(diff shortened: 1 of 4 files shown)


vmfunc:
oh this batch is tight. the negative html matcher plus the `[^'#]` guard on the magento regex so `#env.CRYPT_KEY#` placeholders don't get surfaced as the crypt key, that's exactly the false-positive class that makes config-exposure modules noisy. in it goes once ci's green 🖤

PR #212, opened 2026-06-21
feat(modules): add docker, kubernetes and kubelet api exposure modules

TBX3D:
`modules/recon/docker-api-exposure.yaml` flags an unauthenticated docker engine api on its
`/version` endpoint, keyed on the `ApiVersion` and `MinAPIVersion` fields a generic version
response does not carry, then extracts the engine version. `modules/recon/kubernetes-api-exposure.yaml`
flags an internet-reachable kubernetes api server through its anonymous `/version`, keyed on
`gitVersion` paired with a build field (`gitTreeState`, `buildDate` or `compiler`), then extracts
the git version. `modules/recon/kubelet-api-exposure.yaml` flags an exposed kubelet whose `/pods`
answers without auth, keyed on a `PodList` kind paired with `apiVersion`, then extracts a workload
namespace.

### DIFF
--- a/internal/modules/runtime_api_exposure_test.go
+++ b/internal/modules/runtime_api_exposure_test.go
@@ -1,0 +1,175 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
+:   █▀ █ █▀▀   ·   Blazing-fast pentesting suite                                :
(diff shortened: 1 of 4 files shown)


vmfunc (changes requested):
runtime_api_exposure_test.go is missing the banner, opens straight on `package`. every other test in internal/modules (executor_test.go, matchers_test.go) leads with the ascii block. check-headers skips _test.go so it didn't yell, but it should still have it. add it and this is good, rest of the go's clean.

TBX3D:

vmfunc:
banner's in, merged. docker/k8s/kubelet api detection's good 🖤

PR #213, opened 2026-06-21
feat(modules): add solr, spark and hadoop yarn api exposure modules

TBX3D:
`modules/recon/solr-api-exposure.yaml` flags an exposed apache solr admin api on
`/solr/admin/info/system`, keyed on the `solr-spec-version` paired with the `solr_home`
field a generic json endpoint does not carry, then extracts the solr version.
`modules/recon/spark-api-exposure.yaml` flags an apache spark master whose cluster state
answers without auth on `/json/`, keyed on a `spark://` master url paired with the
`aliveworkers` count (so a non-spark url behind the same count is not reported), then
extracts the master url. `modules/recon/hadoop-yarn-api-exposure.yaml` flags an exposed
hadoop yarn resource manager on `/ws/v1/cluster/info`, keyed on the `clusterInfo` wrapper
paired with `resourceManagerVersion`, then extracts the hadoop version.

### DIFF
--- a/internal/modules/bigdata_api_exposure_test.go
+++ b/internal/modules/bigdata_api_exposure_test.go
@@ -1,0 +1,156 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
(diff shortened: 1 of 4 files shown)


vmfunc:
these are tight. the two-signal AND on each (esp the spark `spark://` url regex + aliveworkers) is exactly what keeps them off generic json, and the negative tests prove it. holding for CI to go green then it's in. nice batch 🖤

PR #214, opened 2026-06-22
feat(modules): add airflow, flink and kafka connect exposure modules

TBX3D:
`modules/recon/airflow-api-exposure.yaml` flags an exposed apache airflow webserver through
its unauthenticated `/health`, keyed on the `metadatabase` and `scheduler` health blocks a
generic health response does not carry, then extracts the scheduler heartbeat.
`modules/recon/flink-api-exposure.yaml` flags an apache flink dashboard reachable without auth
on `/overview`, keyed on the `flink-version` paired with the `slots-total` field a generic
overview lacks, then extracts the flink version. `modules/recon/kafka-connect-api-exposure.yaml`
flags an exposed kafka connect rest api on `/`, keyed on the `kafka_cluster_id` paired with the
`version`, then extracts the version.

### DIFF
--- a/internal/modules/data_pipeline_api_exposure_test.go
+++ b/internal/modules/data_pipeline_api_exposure_test.go
@@ -1,0 +1,151 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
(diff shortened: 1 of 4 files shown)


vmfunc:
clean batch. splitting each service into status + two separate word matchers leans on the default AND so a bare metadatabase or a stray "version" key won't fire, and the negative cases prove it. kafka keying on generic "version" gated by kafka_cluster_id is the right call. in it goes 🖤

PR #215, opened 2026-06-22
feat(modules): add kong, jolokia and nats api exposure modules

TBX3D:
`modules/recon/kong-api-exposure.yaml` flags an exposed kong admin api that grants full control
of the gateway, keyed on the available-plugins map paired with the admin listen address the node
reports, then extracts the kong version. `modules/recon/jolokia-api-exposure.yaml` flags an
exposed jolokia agent bridging http to jmx, keyed on the `agent` and `protocol` fields of its
version response, then extracts the agent version. `modules/recon/nats-api-exposure.yaml` flags
an exposed nats monitoring endpoint that leaks the server topology, keyed on the server id paired
with the max payload, then extracts the server version.

### DIFF
--- a/internal/modules/management_api_exposure_test.go
+++ b/internal/modules/management_api_exposure_test.go
@@ -1,0 +1,155 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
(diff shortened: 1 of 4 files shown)


vmfunc:
these are tight. splitting each fingerprint into separate word matchers so they AND is exactly right, and the half-fingerprint negatives (admin_listen alone, agent without protocol, server_id without max_payload) are what i'd want to see. waiting on the go checks to go green then it's in.

PR #216, opened 2026-06-22
feat(modules): add riak, couchbase and druid api exposure modules

TBX3D:
`modules/recon/riak-api-exposure.yaml` flags an exposed riak http api reachable without auth on
`/stats`, keyed on the `riak_kv_version` paired with the `riak_core_version` a generic stats
endpoint does not carry, then extracts the kv version. `modules/recon/couchbase-api-exposure.yaml`
flags an exposed couchbase cluster management api on `/pools`, keyed on the `implementationVersion`
paired with the `componentsVersion` the bootstrap pool reports, then extracts the implementation
version. `modules/recon/druid-api-exposure.yaml` flags an apache druid process running without
auth on `/status`, keyed on the `org.apache.druid` package namespace paired with the process
`memory` block, then extracts the druid version.

### DIFF
--- a/internal/modules/distributed_db_exposure_test.go
+++ b/internal/modules/distributed_db_exposure_test.go
@@ -1,0 +1,159 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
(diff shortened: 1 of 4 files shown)


vmfunc:
oh these are tight, requiring both kv+core markers on riak (and the impl+components pair on couchbase) kills the false positives nicely, and the negative cases prove it. in once ci clears 🖤

PR #217, opened 2026-06-22
feat(modules): add netdata and cadvisor exposure modules

TBX3D:
`modules/recon/netdata-api-exposure.yaml` flags an exposed netdata agent through its
unauthenticated `/api/v1/info`, keyed on the `mirrored_hosts` and `cores_total` fields a generic
info response does not carry, then extracts the agent version. `modules/recon/cadvisor-api-exposure.yaml`
flags an exposed cadvisor container monitor through its `/api/v1.3/machine` endpoint, keyed on the
`machine_id` and `cpu_frequency_khz` fields, then extracts the machine id.

build/vet/lint clean, `go test ./internal/modules/` green (both modules end to end via

### DIFF
--- a/internal/modules/metrics_exposure_test.go
+++ b/internal/modules/metrics_exposure_test.go
@@ -1,0 +1,138 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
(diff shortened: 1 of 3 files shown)


vmfunc (changes requested):
test file's missing the banner. header-check CI skips *_test.go so it won't yell, but every .go file gets the block. drop the standard banner at the top of internal/modules/metrics_exposure_test.go like matchers_test.go has it. rest of the file is clean.

TBX3D:

vmfunc:
fixed, thanks. netdata + cadvisor, in it goes

PR #218, opened 2026-06-22
feat(modules): add docker registry and harbor exposure modules

TBX3D:
`modules/recon/docker-registry-api-exposure.yaml` flags a docker registry reachable anonymously
through its `/v2/` base, keyed on a 200 paired with the `Docker-Distribution-Api-Version: registry/2.0`
response header (the header rides on a 401 too, so the 200 gate is what proves anonymous reach),
then extracts the api version. `modules/recon/harbor-api-exposure.yaml` flags an exposed harbor
registry through its unauthenticated `/api/v2.0/systeminfo`, keyed on the `harbor_version` and
`auth_mode` fields, then extracts the harbor version.

build/vet/lint clean, `go test ./internal/modules/` green (both modules end to end via

### DIFF
--- a/internal/modules/registry_exposure_test.go
+++ b/internal/modules/registry_exposure_test.go
@@ -1,0 +1,118 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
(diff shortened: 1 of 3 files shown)


vmfunc:
these are tight. matchers being AND-combined means harbor needs both harbor_version and auth_mode so no false-positive on a bare version string, and the 401 negative test is exactly the case i'd worry about. in it goes once ci clears.

PR #219, opened 2026-06-22
feat(modules): add argo cd exposure module

TBX3D:
`modules/recon/argocd-api-exposure.yaml` flags an exposed argo cd api server through its
unauthenticated `/api/version`, keyed on the `KustomizeVersion` and `HelmVersion` fields a generic
version response does not carry, then extracts the server version.

build/vet/lint clean, `go test ./internal/modules/` green (the module end to end via

### DIFF
--- a/internal/modules/argocd_exposure_test.go
+++ b/internal/modules/argocd_exposure_test.go
@@ -1,0 +1,92 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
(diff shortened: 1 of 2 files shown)


vmfunc:
clean module, and requiring both kustomizeversion and helmversion is exactly right, a bare version key was always gonna match half the world. tests cover the helm-alone and kustomize-alone gaps too, nice. holding for ci then it's in.

PR #220, opened 2026-06-22
feat(modules): add metabase, zeppelin and jupyter exposure modules

TBX3D:
`modules/recon/metabase-api-exposure.yaml` flags a metabase instance exposing a live
`setup-token` without auth on `/api/session/properties`, keyed on a non-null token uuid paired
with the `anon-tracking-enabled` setting, then extracts the version tag; a live token is the
pre-auth chain behind `CVE-2023-38646`, and a patched instance reports the token as `null` and is
left alone. `modules/recon/zeppelin-api-exposure.yaml` flags an apache zeppelin server disclosing
its version and build commit over the anonymous `/api/version`, keyed on the `Zeppelin version`
banner paired with the `git-commit-id`, then extracts the version (a version leak rather than an
auth bypass, since the endpoint stays anonymous even on a shiro-secured instance).
`modules/recon/jupyter-api-exposure.yaml` flags a jupyter server whose `/api/status` answers
without a token, keyed on the `last_activity`, `connections` and `kernels` fields, then extracts
the running kernel count.

### DIFF
--- a/internal/modules/analytics_ui_exposure_test.go
+++ b/internal/modules/analytics_ui_exposure_test.go
@@ -1,0 +1,164 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+func runAnalyticsModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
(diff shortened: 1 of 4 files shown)


vmfunc:
clean batch. the metabase setup-token + anon-tracking pairing is the right way to avoid firing on a patched instance, and the null-token negative test proves the uuid regex actually gates it. zeppelin/jupyter matchers are tight too. in it goes 🖤

PR #221, opened 2026-06-22
feat(modules): add influxdb, arangodb and neo4j exposure modules

TBX3D:
`modules/recon/influxdb-api-exposure.yaml` flags an exposed influxdb instance through its
unauthenticated `/health`, keyed on the `influxdb` name paired with the `ready for queries and
writes` health message, then extracts the version. `modules/recon/arangodb-api-exposure.yaml`
flags an arangodb instance reachable anonymously on `/_api/version`, keyed on the `arango` server
name paired with the `version` field, then extracts the version; the 200 gate is what proves
anonymous reach, since an auth-enabled instance answers with a 401.
`modules/recon/neo4j-api-exposure.yaml` flags an exposed neo4j instance through its unauthenticated
root discovery endpoint, keyed on the `neo4j_version` paired with the `neo4j_edition`, then extracts
the version.

### DIFF
--- a/internal/modules/http_database_exposure_test.go
+++ b/internal/modules/http_database_exposure_test.go
@@ -1,0 +1,168 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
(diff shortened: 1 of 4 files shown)


vmfunc:
clean batch. splitting the discriminators into separate word matchers so a bare "version" or "arango" token can't false-positive is exactly the right call, and the 401/404 negatives prove it out. in it goes once ci's green 🖤

PR #222, opened 2026-06-22
feat(modules): add qdrant, weaviate and chroma exposure modules

TBX3D:
`modules/recon/qdrant-api-exposure.yaml` flags a qdrant vector database serving its collections
without an api key on `/collections`, keyed on the `result` envelope wrapping a `collections`
array paired with the `"status":"ok"`, then extracts the first collection name; qdrant gates
`/collections` behind the api key, so an answer here means the catalog is readable.
`modules/recon/weaviate-api-exposure.yaml` flags a weaviate vector database leaking its host
address and version over `/v1/meta`, keyed on a url-valued `hostname` paired with the `version`
field, then extracts the hostname (weaviate drops the modules field when none are enabled, so the
match leans on the always-rendered scheme and host). `modules/recon/chroma-api-exposure.yaml` flags
a reachable chroma vector database by its `/api/v1/heartbeat` and `/api/v2/heartbeat`, keyed on the
`nanosecond heartbeat` field, a reachability signal rather than an auth bypass since the heartbeat
stays anonymous by design.

### DIFF
--- a/internal/modules/vector_db_exposure_test.go
+++ b/internal/modules/vector_db_exposure_test.go
@@ -1,0 +1,143 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dropalldatabases/sif/internal/modules"
+)
+
+func runVectorDBModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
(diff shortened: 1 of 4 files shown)


vmfunc:
these are tight. the AND-across-matchers gating is exactly right, collections-result-but-no-"ok"-status and bare-hostname both correctly fall through, and the negative cases prove it. in it goes once ci's green 🖤

PR #223, opened 2026-06-22
fix(scan): skip built-in module when its legacy flag already ran

TBX3D:
four built-in go modules (nuclei-scan, framework-detection, shodan-lookup,
whois-lookup) wrap a legacy per-target scan that also has its own flag, so
running -am alongside one of those flags (e.g. -am -nuclei) ran the same
per-target scan twice. skip the wrapper module in the run loop when its flag
already ran. securitytrails-lookup is not deduped: its -securitytrails flag
expands the target list rather than scanning a target, so the module is not a
duplicate of it.

### DIFF
--- a/sif.go
+++ b/sif.go
@@ -647,6 +647,24 @@ func (app *App) Run() error {
 				}
 
 				for _, m := range toRun {
+					switch m.Info().ID {
+					case "nuclei-scan":
+						if app.settings.Nuclei {
+							continue
+						}
+					case "framework-detection":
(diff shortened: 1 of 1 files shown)


vmfunc:
yep, running --nuclei alongside --all-modules was double-firing nuclei (and the same for framework/shodan/whois). guard's in the right spot and the case ids match the builtin ids exactly. in it goes. only thing i'd add later is a quick table test for the both-set path so nobody regresses it.

PR #224, opened 2026-06-22
fix(scan): de-duplicate selected modules before running them

TBX3D:
-mt builds the run list by appending modules.ByTag per tag and -m by appending
each id, neither de-duplicated, so a module matching two requested tags
(nuclei-scan is tagged both vuln and cve, so -mt vuln,cve) or listed twice ran
twice. de-duplicate toRun by module id, preserving first-seen order, before
executing.

### DIFF
--- a/sif.go
+++ b/sif.go
@@ -639,6 +639,16 @@ func (app *App) Run() error {
 					}
 				}
 
+				seen := make(map[string]bool, len(toRun))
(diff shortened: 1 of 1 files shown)


vmfunc:
yep, multi-tag modules and dup --modules ids were getting executed twice, this collapses it on Info().ID with first-seen order preserved. clean, in it goes

PR #227, opened 2026-06-23
feat(httpx): retry 429 and 503 with Retry-After backoff

TBX3D:
the shared transport paced outbound requests but ignored a server
asking it to slow down, so a 429 or 503 came straight back as a failed
probe. back off and retry both through the one chokepoint every scanner
shares, honoring Retry-After (delta-seconds or http-date) and falling
back to capped exponential backoff.

gated behind -max-retries (default 2, 0 disables). bodyless GETs and
GetBody-backed requests replay safely; anything else is returned as-is.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -80,6 +80,7 @@ type Settings struct {
 	Header            goflags.StringSlice // custom request headers ("Key: Value")
 	Cookie            string
 	RateLimit         int
+	MaxRetries        int    // -max-retries: retries on 429/503 (0 = off)
 	Notify            bool   // -notify: ship findings to configured providers
 	NotifySeverity    string // -notify-severity: minimum severity to send (info..critical)
 	NotifyConfig      string // -notify-config: path to a notify-compatible yaml file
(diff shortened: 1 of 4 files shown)


vmfunc (commented):
nice work, this is a clean first PR and honestly one of the more carefully defended retry implementations i've reviewed. build, vet, `-race`, and the full suite all pass, gofmt clean.

i spent most of the time on the hostile-input surface, since a `Retry-After` handler is exactly where a server gets to decide how long you sleep, and you covered it:

- `retryAfter` routes every branch through `capDuration` clamping to `[0, 20s]`, so a `Retry-After: 999999` or a far-future http-date can't park a worker. garbage falls through to backoff instead of erroring, and a huge decimal string errors at `strconv.Atoi` before it can overflow. good.
- the `retryBackoffMaxShift = 16` clamp keeps `retryBackoffBase << shift` from overflowing int64, and the cap bounds it regardless. belt and suspenders, i like it.
- `sleepCtx` selects on `ctx.Done()`, and moving `limiter.Wait` inside the loop means every real attempt honors the context too. a cancelled scan won't sleep past its deadline.
- `DrainClose(resp)` before each retry means no body/conn leak between attempts, and the `!rewind` fallback hands the live response back untouched. correct.
- `rewind` gating on `GetBody` so only replayable bodies get resent, and `TestRetryReplaysRequestBody` proves the POST body survives identically. right call.

the tests are real, not decoration: httptest with atomic hit counts asserting exact attempt numbers, both `Retry-After` forms (delta-seconds and http-date), garbage rejection, the cap, the backoff progression, and ctx cancellation. that's the coverage i'd have asked for.

no blockers.

two small things, neither gating:

- the loop retries any method on 429/503, not just idempotent ones, as long as the body rewinds. it's defensible since both statuses mean "i didn't process this, come back," and sif is basically all GET anyway, so i'm fine with it. i'd just drop a one-line comment near the `rewind` call noting that's intentional so the next reader doesn't wonder.
- you unit-test `sleepCtx` cancellation directly, which is enough, but there's no end-to-end test that cancelling mid-retry aborts a real `Do()`. not asking for it, just noting the gap.

pulling it in. thanks for the care on the bounds, that's the part people usually get wrong.

PR #229, opened 2026-06-23
fix(modules): match favicon test to the matchers-condition signature

TBX3D:
#189 gave checkMatchers a condition parameter; the favicon test added in
#183 still called it with the old three-argument form, so internal/modules
fails to compile on main. pass the empty condition the other matcher tests
already use.


TBX3D:
duplicate of #228, which already merged this exact one-liner. closing.

PR #231, opened 2026-06-25
docs(readme): fix invalid discord shield invite link

0x4bs3nt:
- Replaced the invalid Discord invite link badge to the correct one from the top of the README

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,7 @@ If you like the project, but don't have time to contribute, that's okay too! Her
 
 ## Reporting issues
 
-If you believe you've found a bug, or you have a new feature to request, please hop on the [Discord server](https://discord.com/invite/sifcli) first to discuss it.  
+If you believe you've found a bug, or you have a new feature to request, please hop on the [Discord server](https://discord.gg/Yksy9J2BvE) first to discuss it.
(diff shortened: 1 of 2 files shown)


0x4bs3nt:
The claude-review check fails on fork PRs. Not much I can do about this.

vmfunc (commented):
thanks for catching this. the sifcli vanity on that badge was the odd one out, the other discord badge in the readme already points at the Yksy9J2BvE invite, so this makes the two consistent and the syntax stays clean. good fix.

one small follow-up, not blocking this: CONTRIBUTING.md still links discord.com/invite/sifcli, so if the vanity is dead that one should get the same treatment. happy to take it here or in a separate pr.

0x4bs3nt:
> thanks for catching this. the sifcli vanity on that badge was the odd one out, the other discord badge in the readme already points at the Yksy9J2BvE invite, so this makes the two consistent and the syntax stays clean. good fix.
> 
> one small follow-up, not blocking this: CONTRIBUTING.md still links discord.com/invite/sifcli, so if the vanity is dead that one should get the same treatment. happy to take it here or in a separate pr.
> 

i'll push a commit in this PR, give me a few mins.

0x4bs3nt:
@vmfunc fixed :3

PR #233, opened 2026-06-26
feat(modules): detect exposed openai-compatible llm servers

TBX3D:
add recon modules for unauthenticated inference servers that expose an openai-style /v1/models inventory without a key: vllm, llama.cpp, lm studio, and infinity.

### DIFF
--- a/internal/modules/llm_openai_compat_exposure_test.go
+++ b/internal/modules/llm_openai_compat_exposure_test.go
@@ -1,0 +1,225 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runOpenAICompatModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
+	res, err := modules.ExecuteHTTPModule(context.Background(), srv.URL, def, modules.Options{
+		Timeout: 5 * time.Second,
+		Threads: 2,
+	})
+	if err != nil {
+		t.Fatalf("execute %s: %v", file, err)
+	}
+	return res
+}
+
+func openAICompatExtract(res *modules.Result, key string) string {
+	for _, f := range res.Findings {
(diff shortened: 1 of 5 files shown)


vmfunc (commented):
thanks for this, TBX3D. genuinely nice work, the module design here is the good kind of paranoid and it shows.

i pulled just the four yaml modules onto clean main, built ./cmd/sif, and ran them through an adversarial false-positive harness. the headline: the modules are solid. the one thing blocking merge is a test-file collision, not the detections.

what i checked and liked:
- all four are type:http, no size matcher, regex-only extractors, RE2-valid regex, {{BaseURL}} paths, severity:medium, ids kebab-case and unique across modules/. no dead engine features.
- the real risk with openai-compatible servers is that /v1/models looks identical across vllm/llamacpp/lmstudio/infinity/ollama/one-api, so a bare /v1/models + `"object":"list"` match false-positives on everything. you did not do that. each module ANDs a product-unique token: vllm/llamacpp/infinity key off `"owned_by":"vllm"` / `"llamacpp"` / `"infinity"`, and lmstudio smartly uses the native /api/v0/models path plus compatibility_type + quantization + max_context_length, which aren't in the openai schema at all. that's exactly the disambiguation i'd want.
- i served each module a DIFFERENT product's /v1/models list (owned_by openai/library), a catch-all `<html>hi</html>` 200, a 401, and a 404, plus every other module's real body. zero cross-fire. every module still fired on its own real response. that's the whole ballgame for this class of module and it passed.

the blocker (merge, not quality):
internal/modules/llm_inference_exposure_test.go already exists on main from #234 (merged). your PR adds the same path, and both versions define `TestLLMInferenceExposureModules`, `runInferenceExposureModule`, and `inferenceExtract` in package modules_test. that's an add/add conflict, and even if it merged it wouldn't compile (duplicate funcs). #234's copy tests tei/tgi; yours tests vllm/llamacpp/lmstudio/infinity, so they're meant to coexist. rebase onto main and rename your file (something like llm_openai_compat_exposure_test.go) plus those three funcs to unique names. once that's done the test sits next to #234's cleanly.

small nits, none blocking, take or leave:
- infinity lists both /models and /v1/models. a host that serves both gives you two findings for one target. fine if intentional, just noting.
- vllm's regex `"owned_by"\s*:\s*"vllm` has no closing quote so it also matches e.g. vllm-foo. llamacpp and infinity anchor with a trailing `"`. worth tightening to `"vllm"` for consistency.
- the word:"data" matcher in vllm/llamacpp isn't pulling weight, the owned_by regex already carries the match. drop it or make it meaningful.

fix the test-file collision and this is a merge. good stuff.

TBX3D:
ping!

vmfunc:
clean rebase and rename, and you tightened the vllm `owned_by` regex too. builds and tests green on main, no cross-fire with the sibling modules. merging, thanks for the quick turnaround.

PR #234, opened 2026-06-26
feat(modules): detect exposed huggingface inference servers

TBX3D:
add recon modules for unauthenticated text-generation-inference (tgi) and text-embeddings-inference (tei) servers that leak model info without a key.

### DIFF
--- a/internal/modules/llm_inference_exposure_test.go
+++ b/internal/modules/llm_inference_exposure_test.go
@@ -1,0 +1,125 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runInferenceExposureModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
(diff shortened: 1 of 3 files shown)


vmfunc (commented):
this is clean work, thanks for it. both modules are `type: http` and only use status/word/regex matchers + a `type: regex` extractor, so everything actually fires through the engine (no dead size/kval/json paths), and the tests carry real negatives, not just a happy path.

verified:
- `modules/recon/tgi-api-exposure.yaml`: status 200 + regex `"router"\s*:\s*"text-generation-router"` + word `"max_concurrent_requests"`. the router regex is product-unique, so this is well gated.
- `modules/recon/tei-api-exposure.yaml`: status 200 + word `"model_type"` + word `"max_batch_tokens"`. `model_type` alone shows up in a plain HF `config.json`, but requiring `max_batch_tokens` alongside it keeps that from firing, and your `config.json` negative test proves it.
- matchers are AND across the whole list here, which is the right call: these are corroborating signals that genuinely co-occur in a real `/info` payload, not an OR-in-one-list situation.
- regexes are RE2-clean, ids are kebab-case and unique in `modules/`, extractors pull `model_id` on group 1. cross-match negatives (tgi body vs tei module and vice versa) stay silent.

one tiny thing, totally optional:
- `info.author: sif` matches what every existing module in the tree uses, so it's fine as-is. if you'd rather have `TBX3D` on there for credit, say the word and i'll swap it on merge.

heads up unrelated to your code: PRs #235 and #236 add the exact same `internal/modules/llm_inference_exposure_test.go`, so whichever of the three lands first will make the other two rebase. nothing for you to do, just so it's not a surprise.

looks solid. pulling it in.

PR #235, opened 2026-06-26
feat(modules): detect exposed local llm runners

TBX3D:
add recon modules for unauthenticated local inference servers that leak model inventory without a key: ollama, koboldcpp, text-generation-webui (oobabooga), and tabby.

### DIFF
--- a/internal/modules/llm_local_runner_exposure_test.go
+++ b/internal/modules/llm_local_runner_exposure_test.go
@@ -1,0 +1,212 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runLocalRunnerModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
+	res, err := modules.ExecuteHTTPModule(context.Background(), srv.URL, def, modules.Options{
+		Timeout: 5 * time.Second,
+		Threads: 2,
+	})
+	if err != nil {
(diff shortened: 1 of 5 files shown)


vmfunc (commented):
this is a clean set, thanks. four local-runner recon modules with fingerprints that actually discriminate, and you shipped `internal/modules/llm_inference_exposure_test.go` with 18 subtests including the negatives (bare model list, wrong result string, missing-token bodies, plain 200, 404) and an idle-instance positive. that's exactly the shape i want new modules to land in.

went through the engine-level stuff and it all holds up:
- all four are `type: http` so they actually execute (tcp/dns are dead stubs, not a trap you fell into)
- no `size` matchers anywhere, extractors are all `type: regex` with valid group indices, so nothing silently no-ops
- every regex is RE2-clean, no lookahead/backrefs
- the separate-matcher AND layout is the right call here, you want *all* the tokens co-present as a fingerprint, not an OR
- ids are kebab-case and unique, no collision with the existing tree

and the fingerprints aren't the lazy `status: 200` + one generic word that i've been ripping out lately. each one pins a product-specific path plus 2-3 unique co-present tokens (`quantization_level` for ollama, the literal `"result":"KoboldCpp"` + `protected` + `txt2img` for kobold, `model_name` + `lora_names` for ooba, `cpu_info` + `cuda_devices` + `git_describe` for tabby). a catch-all app returning 200 on every path can't fake those. good.

one nit, non-blocking:
- `llm_inference_exposure_test.go` is missing the sif ascii banner the other `_test.go` files carry. header-check CI skips test files so it won't fail anything, it's just cosmetic consistency. add it if you feel like it, i won't hold the merge on it.

(ignore any note about `author: sif`, that's the house convention, every module in the tree uses it, you did it right.)

heads up on my end, not yours: the same test file lands verbatim in a couple sibling PRs, so i'll sequence the merges so nobody has to fight a conflict.

looks solid, pulling it in.

vmfunc:
heads up: i just merged #234, which lands `internal/modules/llm_inference_exposure_test.go`, and #236 adds that same file too. so this now collides on both the file path and the `TestLLMInferenceExposureModules` func name. the four yaml modules here are good to go, it's just the shared test file. can you rebase onto latest main and move your tests into a uniquely-named file + func (e.g. `llm_inference_runners_test.go` / `TestLLMRunnerExposureModules`)? ping me once it's rebased and i'll pull it in.

TBX3D:
ping!

vmfunc:
rebased and renamed cleanly (llm_local_runner), tests green on main. merging, thanks.

PR #236, opened 2026-06-26
feat(modules): detect exposed gpu inference servers

TBX3D:
add recon modules for unauthenticated gpu serving stacks that leak model info without a key: triton, sglang, and torchserve.

### DIFF
--- a/internal/modules/llm_gpu_serving_exposure_test.go
+++ b/internal/modules/llm_gpu_serving_exposure_test.go
@@ -1,0 +1,161 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runGPUServingModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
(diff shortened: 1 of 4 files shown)


vmfunc (commented):
solid work, this is exactly the shape a fingerprint module should be. specific path + two product-unique json tokens ANDed together, plus a real negative test bank. i pulled the engine over it and everything lines up.

what i checked:
- all three are `type: http` so they actually execute (dns/tcp would be dead stubs)
- matchers AND across the whole list, which is what you want here: `triton-api-exposure.yaml` fires only on `status 200` + regex `"name":"triton"` + word `"extensions"`, `sglang-api-exposure.yaml` needs both `"is_generation"` and `"has_image_understanding"`, `torchserve-api-exposure.yaml` needs `"nextPageToken"` and `"modelUrl"`. no single-token status-200 FP bait anywhere.
- regexes are all RE2-clean, extractors are `type: regex` group 1 with valid captures (the only implemented extractor kind, so no silent kval/json no-ops)
- no size matchers, no duplicate ids across `modules/`
- the negative subtests are the part i liked most: hf config, is_generation-alone, modelUrl-without-nextPageToken, ollama list, non-triton kserve, plain 200, 404. that is the anti-FP discipline i want on these.

one tidy-up, non-blocking:
- `author: sif` on all three yamls, should be your handle not the project. i can flip that on merge or you can push it, either way.

heads up (not on you): `internal/modules/llm_inference_exposure_test.go` collides with the same file in #234/#235, so whichever of the three lands first, the others need a quick rebase on that file.

nothing here blocks. i'll fix the author field and merge it in.

vmfunc:
heads up: i merged #234, which lands `internal/modules/llm_inference_exposure_test.go` (same file + `TestLLMInferenceExposureModules` func this pr adds), and #235 is in the same boat. the three yaml modules are good. rebase onto latest main and give your test a unique file + func name (e.g. `llm_inference_serving_test.go` / `TestLLMServingExposureModules`), then ping me and i'll merge.

TBX3D:
ping!

vmfunc:
clean rebase + rename (llm_gpu_serving), green on main. merging, thanks.

PR #237, opened 2026-06-26
feat(modules): detect exposed llm chat frontends

TBX3D:
add recon modules for unauthenticated chat uis that leak config or allow open signup without a key: open webui, librechat, anythingllm, and nextchat.

### DIFF
--- a/internal/modules/llm_ui_exposure_test.go
+++ b/internal/modules/llm_ui_exposure_test.go
@@ -1,0 +1,204 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runUIExposureModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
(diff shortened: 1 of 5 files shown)


vmfunc (commented):
this is a really clean first PR, nice work. the llm frontend exposure angle is a good addition and the test file is genuinely thorough, the per-module negative cases (auth-enabled, closed-registration, code-gated nextchat, missing-key-per-module, plain 200, 404) are exactly what i want to see on a new detection module.

verified locally:
- all four are `type: http` so they actually execute, matchers combine correctly (each module wants ALL its product tokens present, so separate AND-combined matchers is the right call here, not an OR case)
- every extractor is `type: regex` with `group: 1`, all patterns RE2-clean, no `size`/`kval`/`json` traps
- ids are kebab-case and globally unique, no collisions
- `go build ./cmd/sif` + `go vet` clean, `TestLLMUIExposureModules` passes 17/17
- FP behavior is solid: you key on multiple product-unique json tokens rather than a bare `status:200`, so the catch-all/plain-200/404 cases stay silent. the three modules sharing `{{BaseURL}}/api/config` don't cross-fire since their token sets are disjoint.

one tiny thing, not blocking:
- `author: sif` on all four yamls should be your handle (`TBX3D`), that field is meant to credit the contributor. fix it if you push again, otherwise i'll just tweak it on my end.

couple of notes for later, no action needed:
- `open-webui-exposure` at medium is defensible but honestly leans high, it's a fully unauth chat ui with the operator's model connections reachable. fine to leave as is.
- `anythingllm-exposure` keys on the config disclosure (LLMProvider/VectorDB/EmbeddingEngine) rather than the `RequiresAuth:false` field, which is a valid info-disclosure signature, just flagging endpoint reachability more than auth-off state. totally fine.

merging it in. thanks for the solid contribution.

PR #238, opened 2026-06-26
feat(modules): detect exposed rag and knowledge apps

TBX3D:
add recon modules for unauthenticated rag/search apps that leak config or health without a key: perplexica, verba, onyx, and fastgpt.

### DIFF
--- a/internal/modules/llm_rag_ui_exposure_test.go
+++ b/internal/modules/llm_rag_ui_exposure_test.go
@@ -1,0 +1,201 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runRAGUIModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
+	res, err := modules.ExecuteHTTPModule(context.Background(), srv.URL, def, modules.Options{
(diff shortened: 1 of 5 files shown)


vmfunc (commented):
this is really clean work, thanks for the thorough test coverage. all four modules are `type: http`, extractors are all `type: regex`, patterns are RE2-valid, paths use `{{BaseURL}}`, and the ids are unique against the tree. built and ran it on your head commit: `go build ./cmd/sif` is green and all 17 subtests of `TestLLMUIExposureModules` pass.

what i like: these are proper multi-token AND fingerprints on product-specific paths, not `status: 200` + one generic word. since sif ANDs every matcher in the list, dropping any single token silences the module, and your negatives prove exactly that. `verba-health-exposure.yaml` requiring a non-empty `WEAVIATE_URL_VERBA` via the regex matcher so an empty/embedded deployment doesn't fire is a nice touch.

no blockers. a couple of small things:

- `author: sif` on all four yamls, should be you (TBX3D). two-second fix, i can also just patch it on merge if you'd rather.
- `onyx-auth-exposure.yaml` severity=medium is a touch high imo, `/api/auth/type` only leaks auth-config booleans (auth type, anon enabled, password min length, has_users), no secret material. low/info would be defensible. your `high` on perplexica and verba is correct since those genuinely leak provider api keys / weaviate creds.

one heads-up not on you: PR #237 adds `internal/modules/llm_ui_exposure_test.go` too with different contents, so #237 and #238 can't both merge clean on that file. i'll sort the ordering on my side, just flagging so it's not a surprise. (the missing ascii banner on that test file is fine, header-check skips `*_test.go`.)

nothing here blocks it. i'll fix up the author field and pull it in.

vmfunc:
heads up: i merged #237, which lands `internal/modules/llm_ui_exposure_test.go` (same file + `TestLLMUIExposureModules` func this pr adds), so this now conflicts on main. the yaml modules are good to go. rebase onto latest main and rename your test file + func to something unique (e.g. `llm_ui_rag_test.go` / `TestLLMRagUIExposureModules`), ping me and i'll pull it in.

TBX3D:
ping!

vmfunc:
clean rebase + rename (llm_rag_ui), green on main. merging, thanks.

PR #239, opened 2026-06-26
feat(modules): detect exposed data labeling tools

TBX3D:
add recon modules for unauthenticated annotation servers that leak projects or config without a key: cvat and label studio.

### DIFF
--- a/internal/modules/ml_platform_exposure_test.go
+++ b/internal/modules/ml_platform_exposure_test.go
@@ -1,0 +1,133 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runMLPlatformModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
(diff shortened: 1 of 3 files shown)


vmfunc (commented):
nice work, this is clean module work. both `cvat-server-exposure.yaml` and `label-studio-exposure.yaml` are `type: http`, use only the matcher/extractor types the engine actually implements (status/word/regex), and gate on product-unique tokens so they hold up against a catch-all 200 app.

the AND-across-matchers logic is right for what you want here:
- `cvat-server-exposure.yaml`: status 200 AND `"name":"Computer Vision Annotation Tool"` regex AND `"version"` AND `"logo_url"`. the name-regex carries the specificity, so the two generic word matchers are fine under AND.
- `label-studio-exposure.yaml`: status 200 AND `"label-studio-os-package"` AND `"label-studio-os-backend"`. both tokens are product-unique, so no `<html>` negative matcher needed.

ran your `TestMLPlatformExposureModules` plus my own fixtures (catch-all 200, 401 login json, `{}`, prose-mentioning-cvat html): fires on real product json, silent on everything else. regexes are RE2-clean, ids are unique, `low` severity is right for an unauth version/info leak.

`author: sif` is consistent with every other module in the repo, so no change there.

one heads-up, not a blocker on your side: PR #240 also adds `internal/modules/ml_platform_exposure_test.go` with different contents, so these two will collide on that file. whichever of us merges first, the other needs a rebase. i'll sort the ordering when i pull these in.

merging this in.

PR #240, opened 2026-06-26
feat(modules): detect exposed ml platforms

TBX3D:
add recon modules for unauthenticated ml platforms that leak cluster or deployment info without a key: h2o, mindsdb, and zenml.

### DIFF
--- a/internal/modules/ml_platform_server_exposure_test.go
+++ b/internal/modules/ml_platform_server_exposure_test.go
@@ -1,0 +1,161 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runMLPlatformServerModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
+	res, err := modules.ExecuteHTTPModule(context.Background(), srv.URL, def, modules.Options{
+		Timeout: 5 * time.Second,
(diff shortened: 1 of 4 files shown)


vmfunc (commented):
this is a clean set, nice work. all three are `type: http` so they actually execute, and the matcher shape is exactly right: `status:200` plus three separate product-specific json-key `word` matchers, which AND together so you only fire when all the keys are present. that conjunction is doing the fp-guarding, and it holds up: no fires against catch-all html 200s, 401s, empty `{}`, or a SPA index. extractors are `type: regex` (the only implemented kind) and the patterns are RE2-clean. ids are unique and kebab-case. the test file with drop-one-key negatives is genuinely good hygiene, thank you for including it.

no blockers on the module logic itself. one coordination thing before it lands though:

- **file collision with #239**: your `internal/modules/ml_platform_exposure_test.go` and #239's are the same path *and* both declare `func TestMLPlatformExposureModules`, with different bodies (yours h2o/mindsdb/zenml, theirs label-studio/cvat). they'll conflict on merge, and even force-merged the duplicate func won't compile. whichever of these goes second needs the test file + func renamed (e.g. `ml_platform_exposure_h2o_test.go` / `TestH2OMindsDBZenMLExposure`). not a defect in your yaml, just needs sequencing.

nits (non-blocking):
- `modules/recon/h2o-cluster-exposure.yaml` is `severity: high`, but your own description says arbitrary code execution / cluster takeover on a reachable node, so `critical` is defensible. your call.
- no negative matcher excluding generic html like `maven-settings-exposure.yaml` has, but your three unique json keys make it unnecessary here, so leave it.
- `author: sif` is fine, that matches every existing module in the tree.

so: rename the test file/func to avoid the #239 clash and i'll pull it in. everything else is good to go.

vmfunc:
heads up: i merged #239, which lands `internal/modules/ml_platform_exposure_test.go` with a `TestMLPlatformExposureModules` func. yours is the same path + func name with different bodies, so it conflicts now (git won't take both, and even forced they'd be a duplicate-func compile error). the modules are good. rebase onto latest main and rename your test file + func (e.g. `ml_platform_serving_test.go` / `TestMLServingExposureModules`), then ping me and i'll merge.

TBX3D:
ping!

vmfunc:
clean rebase + rename (ml_platform_server), green on main. merging, thanks.

PR #241, opened 2026-06-26
feat(modules): detect exposed ai image generation servers

TBX3D:
add recon modules for self-hosted image generation servers reachable
without auth: comfyui, automatic1111, fooocus-api, and iopaint each
expose unauthenticated generation or editing and disclose the installed
models.

### DIFF
--- a/internal/modules/image_gen_exposure_test.go
+++ b/internal/modules/image_gen_exposure_test.go
@@ -1,0 +1,195 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runImageGenModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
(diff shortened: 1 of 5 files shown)


vmfunc (commented):
nice set. four clean recon modules for self-hosted image-gen servers, and the test file is genuinely thorough, the negatives (comfyui version-without-vram, a1111 partial/titleless entries, fooocus missing either filename key, iopaint missing any of the three config keys) are exactly the cases i'd worry about for FPs. good instinct.

verified the important stuff:
- all four are `type: http` so they actually run (dns/tcp are stubs), paths use `{{BaseURL}}`, extractors are all `type: regex` (the only implemented kind), and the four regexes are RE2-clean.
- the matcher structure is right. each module ANDs `status: 200` with the product-unique tokens on a product-specific path (`/system_stats` + `comfyui_version`+`torch_vram_total`, `/sdapi/v1/sd-models` + `model_name`+`filename`+`title`, `/v1/engines/all-models` + `model_filenames`+`lora_filenames`, `/api/v1/server-config` + `interactiveSegModel`+`enableFileManager`+`disableModelSwitch`). since matchers AND across the whole list, that's the correct intent here and it's what keeps them off catch-all html / login json / empty `{}`.
- ids are globally unique, no collision with the existing set.

one thing to fix before i pull it in:
- `info.author: sif` on all four should be your handle, not `sif`. drop `TBX3D` in there so you get credit.

nits, take or leave:
- `modules/recon/comfyui-api-exposure.yaml` is `high` (fair, unauth workflow + custom-node exec is basically RCE) but `automatic1111-api-exposure.yaml` at `medium` is arguably light given `/sdapi/v1/sd-models` leaks absolute local model paths and the api allows unauth generation. medium is defensible, your call.
- no negative matcher on any of them. not required here since the tokens are specific enough that generic html/login/empty-json stay silent, but if you ever loosen a token, add a `negative` matcher excluding `<html>`/`<!DOCTYPE` like `modules/recon/maven-settings-exposure.yaml` does.

fix the author field and i'll merge it in. solid contribution.

PR #242, opened 2026-06-26
feat(modules): detect exposed ai orchestration platforms

TBX3D:
add recon modules for self-hosted agent builders and compute
orchestration. ray exposes an unauthenticated compute dashboard whose
job api allows code execution, skypilot exposes an open cloud and
kubernetes control plane when basic auth is disabled, dify flags a
console that allows open registration, and langflow fingerprints a
reachable instance over its public version api.

### DIFF
--- a/internal/modules/ai_orchestration_exposure_test.go
+++ b/internal/modules/ai_orchestration_exposure_test.go
@@ -1,0 +1,216 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runOrchestrationModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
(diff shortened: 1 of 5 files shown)


vmfunc (commented):
nice work, this is a clean, well-scoped set. pulled it on top of main, builds fine and `go test ./internal/modules/...` is green, and your ai_orchestration_exposure_test.go is genuinely thorough. i also ran my own adversarial pass (catch-all 200 html/json, 401, 404, empty body, plus every module against the other three products' real responses and a plausible adjacent vendor for each) and nothing false-fires. that cross-product check is the one i usually catch people on, yours holds.

engine correctness, all four:
- type: http, every path is {{BaseURL}}/..., matchers are all status/word/regex, extractors are all type: regex. nothing lands on an unimplemented path.
- matchers are default-AND across the list and you leaned on that correctly: specific path + product-unique tokens ANDed, not a bare status:200. all regex is RE2-clean.
- ids are kebab-case and globally unique, grepped modules/, no collisions.

what i actually like: two of these only fire on the insecure state, not mere presence. dify requires `is_allow_register: true`, skypilot requires `basic_auth_enabled: false`. so a dify with registration closed or a skypilot behind auth won't get flagged, and your negatives prove it. that's the right instinct.

one thing, non-blocking, severity:
- ray (high) and skypilot (high) are right. ray's version endpoint being open means the job submission api is open too = pre-auth rce, high is fair.
- dify at medium undersells it. open registration on an exposed console means anyone provisions a workspace and burns the org's llm creds. i'd bump that to high.
- langflow at medium i get, you scoped it honestly as a presence fingerprint and the version endpoint is public by design. but given langflow's unauth-rce track record a reachable instance is the exploitable surface, so i lean high there too. your call, the description is honest either way so i won't gate on it.

none of that blocks. pulling this in as-is, worth a follow-up to nudge dify to high. thanks for the clean pr and the real test coverage.

PR #243, opened 2026-06-26
feat(modules): detect exposed ml experiment trackers

TBX3D:
add recon modules for self-hosted training and experiment-tracking
platforms reachable without auth: mlflow, tensorboard, aim, and
determined disclose experiments, the artifact store, training run paths,
and cluster topology over unauthenticated apis.

### DIFF
--- a/internal/modules/ml_tracking_exposure_test.go
+++ b/internal/modules/ml_tracking_exposure_test.go
@@ -1,0 +1,223 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runTrackingModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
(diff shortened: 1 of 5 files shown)


vmfunc (commented):
solid work, this is one of the cleaner module PRs i've gotten. good coverage on a real gap (self-hosted ml trackers left open on the internet is genuinely common).

went over it against the engine and it's all correct:
- all four are `type: http`, so they actually run (dns/tcp are stubs, would've been dead)
- matcher structure is `status: 200` + a separate `word` matcher per token, AND'd across the list. that's the right call here, you want *all* the product-unique keys present. e.g. `modules/recon/aim-exposure.yaml` requiring `telemetry_enabled` + `warn_index` + `warn_runs` together, none of which show up on a random 200 page.
- extractors are all `type: regex` (the only implemented kind), patterns are RE2-clean, no lookahead/backrefs
- every path hits a specific endpoint (`/api/2.0/mlflow/experiments/get-by-name`, `/data/environment`, `/api/projects/`, `/api/v1/master`) with `{{BaseURL}}`, not a bare root probe
- ids are kebab-case and unique against the tree

the test file is the part i'm happiest about. `ml_tracking_exposure_test.go` has per-token negative cases (drop any one key and it goes silent) plus plain-200 and 404 negatives across all four. that's exactly the FP discipline i want, since a status-200-only module firing on a catch-all app is the recurring bug here.

no blockers. `author: sif` is fine, that's what the whole existing tree uses.

one optional follow-up if you feel like it, not required: `modules/recon/maven-settings-exposure.yaml` pairs its positive tokens with a `negative: true` matcher excluding `<html>`/`<!DOCTYPE`. your unique JSON keys already make html false-positives basically impossible (confirmed empirically, catch-all html + 401 + `{}` all stay silent), so it's belt-and-suspenders, but it's the house pattern if you want to match it.

pulling it in. thanks.

PR #244, opened 2026-06-26
feat(modules): detect exposed vector search engines

TBX3D:
add recon modules for self-hosted vector and semantic search engines
reachable without auth: marqo, vespa, and meilisearch each allow
unauthenticated read and write of the indexed data, and their root or
version endpoints disclose the build.

### DIFF
--- a/internal/modules/vector_search_exposure_test.go
+++ b/internal/modules/vector_search_exposure_test.go
@@ -1,0 +1,150 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runVectorSearchModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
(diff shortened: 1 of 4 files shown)


vmfunc (commented):
this is a clean set, nice work. three tight recon modules for marqo, vespa and meilisearch, all `type: http`, all using status + word/regex matchers that AND correctly across the list, and every one pins a specific path plus product-unique tokens instead of a bare `status: 200 + generic word`. exactly the shape i want after the FP fixes in 247.

specifics i checked:
- `modules/recon/marqo-exposure.yaml`: `/` + regex `"message"\s*:\s*"Welcome to Marqo"` + `"version"` word. tight, the version-in-prose html case is correctly excluded because it keys on the structured `"message":` shape.
- `modules/recon/meilisearch-exposure.yaml`: `/version` ANDing `commitSha` + `commitDate` + `pkgVersion`. all three have to be present, so a bare `{}` or a login json wont trip it.
- `modules/recon/vespa-status-exposure.yaml`: `/ApplicationStatus` ANDing `abstractComponents` + `processingChains` + `httpRequestFilters`. same story.

all regexes are RE2-clean (no lookahead/backref), extractors are `type: regex` only so nothing hits the unimplemented kval/json path, ids are unique, and `severity: high` is right for unauthenticated read/write on the data plane. your `vector_search_exposure_test.go` earns its keep too, the html-mention and generic-welcome negatives are the exact cases i'd have poked at.

one note, not a change request: the meili and vespa modules use three separate `word` matchers each rather than one `condition: and` word list. works identically since matchers AND across the list, purely a readability thing if you ever touch them again.

and to correct the automated pass: `author: sif` is fine, that's the existing convention across every module in `modules/recon/` (maven/nuget/gradle/backup all use it), so no change needed there.

merging this in. thanks.

PR #245, opened 2026-06-26
feat(modules): detect exposed ai speech servers

TBX3D:
add recon modules for self-hosted speech inference servers that are exposed without authentication: speaches (faster-whisper successor, speech-to-text and text-to-speech) keyed on its non-openai model task field, and xtts-api-server keyed on its get-folders endpoint that discloses local filesystem paths.

### DIFF
--- a/internal/modules/speech_audio_exposure_test.go
+++ b/internal/modules/speech_audio_exposure_test.go
@@ -1,0 +1,143 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runSpeechModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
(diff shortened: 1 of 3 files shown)


vmfunc (commented):
this is a really clean first module PR, honestly. nice work.

went through it against the engine and it holds up:
- both are `type: http` so they actually run (tcp/dns are stubs here), build's green
- no size matcher, no kval/json extractor. the regex extractors use `group: 1` against 1-subexp patterns so they populate correctly (`speaches_model`, `xtts_model_folder`)
- all four regexes compile under RE2, no lookahead/backref
- the AND-across-matchers is the right call in both. in `modules/recon/speaches-api-exposure.yaml` the `"task": "(automatic-speech-recognition|text-to-speech|...)"` regex is what keeps you off a generic openai-compatible `/v1/models`, i confirmed it returns false against a taskless vllm/ollama list, so that's doing exactly the work you want. in `modules/recon/xtts-api-server-exposure.yaml` requiring all three folder keys keeps it off a catch-all 200.
- your test file is genuinely good. the negative cases (taskless list, `fill-mask` task, empty `data`, each-missing-folder-key, plain 200, 404) are exactly what i'd have asked for.

one tiny thing before i pull it in:
- `author: sif` in both yamls should be your handle, not the project. that's the only change i want.

nit, not blocking: severity `medium` is fine, though you could argue `high` given both are fully unauth inference APIs and xtts leaks local fs paths via `get_folders`. your call, i won't push on it.

flip the `author` field to your handle and i'll merge it in.

PR #246, opened 2026-06-26
feat(modules): detect exposed one-api llm gateway

TBX3D:
add a recon module for the one-api / new-api self-hosted llm gateway,
whose public status endpoint discloses the system name, version, and
quota configuration.

### DIFF
--- a/internal/modules/llm_gateway_exposure_test.go
+++ b/internal/modules/llm_gateway_exposure_test.go
@@ -1,0 +1,85 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runGatewayModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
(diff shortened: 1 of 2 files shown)


vmfunc (commented):
nice one, this is a clean module. tested it end to end and it holds up: `go build` green, `TestLLMGatewayExposureModules` passes with the four negatives (missing `system_name`, missing `quota_per_unit`, plain 200, 404), and i ran my own catch-all `<html>` 200-on-every-path check which returns 0 findings. no FPs.

the matcher design is right: the four entries in `modules/recon/oneapi-status-exposure.yaml` AND together (status 200 + `system_name` + `start_time` + `quota_per_unit`), and those three tokens are genuinely co-present in a real one-api/new-api `/api/status` body, so it's specific without being brittle. `oneapi_version` regex extractor is valid RE2 and pulls `v0.8.4` correctly.

one tiny thing, not a blocker:

- `author: sif` should be your handle (`TBX3D`). house convention is author = contributor. i can flip it on merge or you can push a one-liner, your call.

optional hardening for later, don't need it now: a `negative` matcher excluding generic `<html>`/`<!DOCTYPE` like `modules/recon/maven-settings-exposure.yaml` has. empirically unnecessary here since the three tokens already gate it, just belt-and-suspenders against odd catch-all json apps. severity: medium for version+uptime+oauth-config+quota leak on a credential-handling proxy is defensible, leaving it.

pulling this in. thanks 🌱

PR #247, opened 2026-06-26
fix(frameworks): stop false positives and version mis-extraction

TBX3D:
Follows the line of #133 (drop bare-substring signatures that false-positive). A detector-accuracy sweep turned up two classes of bug in the framework detectors, each reproducible against `main`.

## Bare-brand header false positives

`HeaderOnly` signatures match a brand name as a substring across **every** header name and value, so a detector fires on any response that merely references the brand — a vendor CDN named in a `Link`/CSP value, or a cookie that shares the brand prefix.

This adds an optional `Header` field to `Signature` that scopes a header-only match to a single named header's value (backwards compatible; empty keeps the old behaviour), then fixes each detector:

| detector | fired on | fix |
| --- | --- | --- |
| Express | `Set-Cookie: express_checkout=1` | scope `Express` to `X-Powered-By` |
| Flask | a `Link:` to the werkzeug docs | scope `Werkzeug` to `Server` |
| Symfony | a `Link:` to `symfony.com` | drop bare `symfony` (no such header exists), use `X-Debug-Token` |
| Shopify | a `Link:` to `cdn.shopify.com` | key on the `x-shopify-*` response headers |
| Remix | a `track_remix.mp3` asset | drop bare `remix`/`_remix`, keep `__remixContext` |
| Spring Boot | a tutorial discussing the error | anchor `>Whitelabel Error Page<` in its `<h1>` |

The **Gin** and **FastAPI** detectors are removed. Gin keyed on the `gin-gonic` import-path string, which shows up in tutorials and source listings but never in a real Gin response; FastAPI keyed on bare `fastapi`/`starlette` words that match the projects' doc domains. Neither framework sets an identifying response header or emits a non-prose body marker, so there is no clean passive signal to anchor on.

## Version mis-extraction

- The low-confidence `.*?` version fallbacks (`rails`, `django`, `laravel`, `spring`) have an unbounded gap that grabs the first version-shaped number anywhere after the framework word. With no real version present they report an unrelated asset's cache-buster — e.g. `rails` plus a later `/app-7.8.9.css` yields `7.8.9`. Dropped; the anchored high-confidence patterns that require the version adjacent to the name are kept.
- `isValidVersionString` required `dotCount >= 1`, rejecting single-integer versions. Drupal's generator ships a bare major (`Drupal 10`), so its version always came back `unknown`. It now accepts a single integer.

## Tests

New regression tests cover every false positive (fires vs. quiet) and both version bugs, e.g.:

```go
// fires on a real X-Powered-By, quiet on an express_checkout cookie
expressDetector{}.Detect("", http.Header{"X-Powered-By": {"Express"}})           // > 0.5
expressDetector{}.Detect("", http.Header{"Set-Cookie": {"express_checkout=1"}})  // <= 0.5

frameworks.ExtractVersionOptimized(`<link href="/app-7.8.9.css">`, "Ruby on Rails") // not "7.8.9"
frameworks.ExtractVersionOptimized(`content="Drupal 10"`, "Drupal")                 // "10"
```

`gofmt`, `go vet`, `golangci-lint`, `go test -race`, and the integration build all pass.

### DIFF
--- a/internal/scan/frameworks/detect_test.go
+++ b/internal/scan/frameworks/detect_test.go
@@ -770,39 +770,6 @@ func TestDetectFramework_Htmx(t *testing.T) {
 	}
 }
 
-func TestDetectFramework_GinFalsePositive(t *testing.T) {
-	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Access-Control-Allow-Origin", "*")
-		w.WriteHeader(http.StatusOK)
-		w.Write([]byte(`<!DOCTYPE html><html><body>Welcome</body></html>`))
-	}))
-	defer server.Close()
-
-	result, err := frameworks.DetectFramework(server.URL, 5*time.Second, "")
-	if err != nil {
-		t.Fatalf("unexpected error: %v", err)
-	}
-	if result != nil && result.Name == "Gin" {
-		t.Errorf("false positive: detected Gin (confidence %.2f) on a CORS header", result.Confidence)
-	}
-}
-
-func TestDetectFramework_Gin(t *testing.T) {
-	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(http.StatusNotFound)
-		w.Write([]byte(`404 page not found - powered by gin-gonic`))
-	}))
-	defer server.Close()
-
-	result, err := frameworks.DetectFramework(server.URL, 5*time.Second, "")
-	if err != nil {
-		t.Fatalf("unexpected error: %v", err)
-	}
-	if result == nil || result.Name != "Gin" {
-		t.Errorf("expected framework 'Gin', got '%v'", result)
-	}
-}
-
 func TestDetectFramework_MeteorFalsePositive(t *testing.T) {
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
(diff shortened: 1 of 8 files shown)


vmfunc (commented):
this is a good one, thanks. the header-scoping approach is exactly right and the accuracy_test.go / version_test.go pairs (fires vs quiet on every case) are the kind of regression coverage i wish every detector PR shipped with. builds clean, vet clean, `go test ./internal/scan/frameworks/...` green on my end too. no blockers, i'm pulling it in.

two residuals to flag, both non-blocking (they're strictly narrower than what's on main today, so this PR still reduces FPs net):

- `internal/scan/frameworks/detectors/backend.go` symfony sig `{Pattern: "X-Debug-Token", HeaderOnly: true}` is left unscoped, so it falls through to `containsHeader` which scans header names *and* values. a non-symfony origin that lists `X-Debug-Token` in a CORS `Access-Control-Allow-Headers`/`Expose-Headers` value scores 0.731. same class you're fixing, and higher conf than the shopify one below.
- `internal/scan/frameworks/detectors/cms.go` shopify sig `{Pattern: "x-shopify", HeaderOnly: true}` same deal, 0.531 on `X-Shopify-*` header names appearing in a CORS/`Vary` value.

the wrinkle is your new `Header` field scopes to one header's *value*, but symfony/shopify are really "does this header *name* exist" signals across a family of headers, so `Header:` doesn't cleanly fit them the way it does X-Powered-By / Server. fine to leave for a follow-up. if you want to close it here, a small "header name present" check (distinct from the value-scan path) would do it. either way it's not holding up the merge.

nits, take or leave:
- `flask` still has the unscoped bare `{Pattern: "flask", HeaderOnly: true}` sig. a CSP/Link value naming flask.palletsprojects.com lands at exactly 0.500, so it's saved only by the strict `> 0.5` threshold. one weight nudge anywhere and it tips. worth scoping or dropping like you did the others.
- dropping the Shopify/Spring Boot version entries means those always report `unknown` now. intentional and fine given there's no clean passive version signal, just noting it.

merging. nice work.

PR #248, opened 2026-06-26
feat(modules): add loki, jaeger and zipkin exposure modules

TBX3D:
`modules/recon/loki-api-exposure.yaml` flags a grafana loki deployment running with `auth_enabled` off, keyed on the `"status":"success"` envelope paired with the `data` label array on `/loki/api/v1/labels`, then extracts the first label; loki wants the `X-Scope-OrgID` tenant header when auth is on (401 without it), so a tenant-less 200 means logs are readable. `modules/recon/jaeger-query-exposure.yaml` flags an exposed jaeger query service leaking its internal service topology over `/api/services`, keyed on the `data` array paired with the `total`/`limit`/`offset`/`errors` envelope fields, then extracts the first service. `modules/recon/zipkin-exposure.yaml` flags a reachable zipkin tracing server over `/zipkin/config.json`, keyed on the `queryLimit`, `defaultLookback` and `searchEnabled` lens-ui config fields, then extracts the environment.

### DIFF
--- a/internal/modules/observability_exposure_test.go
+++ b/internal/modules/observability_exposure_test.go
@@ -1,0 +1,135 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runObservabilityModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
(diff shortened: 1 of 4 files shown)


vmfunc (commented):
nice work on these, three clean observability modules. specific paths + product-unique json tokens, `condition: and` used correctly, and you shipped real negative tests (401, non-loki envelope, bare data array, partial zipkin config, plain 200, 404). that's exactly the shape i want, and it holds up: matchers are AND across the whole list so the token combos genuinely gate, regexes are RE2-clean, extractors line up on group 1, and no size/kval/json footguns anywhere. ids are unique, and it builds + tests green merged on main.

went through each:
- `modules/recon/loki-api-exposure.yaml`: the `"status":"success"` + `"data"` envelope is the same shape prometheus uses, but the `/loki/api/v1/labels` path carries the specificity and your negatives confirm it doesn't fp. fine as-is.
- `modules/recon/jaeger-query-exposure.yaml` and `modules/recon/zipkin-exposure.yaml`: product-specific paths + unique tokens, no bare `status: 200` bait. good.

only nit: the license banner isn't needed on the `*_test.go` file, header-check skips those, so ignore that one.

no blockers, pulling it in. thanks.

PR #249, opened 2026-06-26
feat(modules): add traefik, nomad and portainer exposure modules

TBX3D:
`modules/recon/traefik-api-exposure.yaml` flags an exposed traefik api over `/api/overview`, keyed on the `features` and `accessLog` blocks paired with the `providers` list, then extracts the first provider; the api is only mounted when insecure mode or the dashboard is enabled, so a 200 means the routing config is readable. `modules/recon/nomad-agent-exposure.yaml` flags a hashicorp nomad agent with acls disabled over `/v1/agent/self`, keyed on the `config`, `member`, `stats` and `Tags` blocks, then extracts the build version; an acl-enabled agent answers 403, so a 200 dumps the full agent configuration. `modules/recon/portainer-status-exposure.yaml` flags a reachable portainer instance over its pre-login `/api/status`, keyed on the `Version` and `InstanceID` fields, then extracts the instance id.

### DIFF
--- a/internal/modules/infra_controlplane_exposure_test.go
+++ b/internal/modules/infra_controlplane_exposure_test.go
@@ -1,0 +1,129 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runInfraControlplaneModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
+	res, err := modules.ExecuteHTTPModule(context.Background(), srv.URL, def, modules.Options{
+		Timeout: 5 * time.Second,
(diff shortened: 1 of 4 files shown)


vmfunc (commented):
nice set, these are clean modules. specific paths, multiple product-unique JSON tokens ANDed together with status 200, and you shipped real negative coverage (403 acl'd nomad, config+stats-without-member, bare version, plain 200, 404). ran the matcher logic against the engine and it holds: no FP on a catch-all 200 html app or a bare `{}`, regexes are all RE2-clean, extractor groups line up. no functional blockers.

couple of things to tighten before i pull it in, all metadata not logic:

- `portainer-status-exposure.yaml` severity `medium` -> drop to `low`/`info`. `/api/status` is an intentional pre-login endpoint and you're only pulling `Version` + `InstanceID`, that's a fingerprint banner, not an exposure. the module itself is good, just score it as recon.
- two descriptions claim more than the matcher proves:
  - nomad says "accepts tokenless job submission" but the module only does `GET /v1/agent/self`, it never touches the jobs api. reword to what you actually detect (acls-disabled agent dumping its config).
  - traefik says "full routing config and internal backend hosts" but `/api/overview` is just the summary (router/service counts + features + provider names). the backend hosts live at `/api/http/routers` / `/api/http/services` which you don't hit. say it exposes the overview / that the api is reachable.

nomad `high` and traefik `high` are correctly scored, an open unauth controlplane is a real pivot.

retune the portainer severity and tighten those two descriptions and i'll merge it in. good work.

TBX3D:
ping!

vmfunc:
all three addressed: portainer's `low` now, and both the nomad and traefik descriptions match what the modules actually probe. green on main. merging, nice work.

PR #250, opened 2026-06-26
feat(modules): add concourse, woodpecker and gocd exposure modules

TBX3D:
`modules/recon/concourse-info-exposure.yaml` flags a concourse ci server over its unauthenticated `/api/v1/info`, keyed on the `version`, `worker_version` and `external_url` fields, then extracts the version. `modules/recon/woodpecker-version-exposure.yaml` flags a woodpecker ci server over `/version`, keyed on the `version` field paired with a `source` url pointing at woodpecker (so a drone server of the same shape stays quiet), then extracts the version. `modules/recon/gocd-version-exposure.yaml` flags a gocd server over `/go/api/version`, keyed on the `full_version` and a `commit_url` under `github.com/gocd/gocd`, then extracts the version.

### DIFF
--- a/internal/modules/cicd_server_exposure_test.go
+++ b/internal/modules/cicd_server_exposure_test.go
@@ -1,0 +1,120 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runCICDModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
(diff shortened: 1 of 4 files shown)


vmfunc (commented):
nice work, these are clean. specific paths, product-unique token combos, and the near-miss tests (missing `worker_version`, drone `source`, non-gocd `commit_url`) show you understood the FP model, which is exactly what i care about here. pulled the branch, build's green, your 8 subtests pass, and my own catch-all-html / generic-json probes all stayed quiet.

no blockers. two small asks before i merge:

- `author: sif` on all three should be your handle instead. convention is the contributor owns the module.
- severity: `gocd-version-exposure.yaml` and `woodpecker-version-exposure.yaml` are bare version banners, those want `low` (or `info`). `concourse-info-exposure.yaml` can stay `medium` since it also leaks `worker_version` + `external_url`.

one nit, take it or leave it: in `woodpecker-version-exposure.yaml` the `"source"\s*:\s*"[^"]*woodpecker` regex is an unanchored substring, so any source url containing "woodpecker" anywhere matches. in practice that's woodpecker or a fork so it's fine, but `woodpecker-ci/woodpecker` would be tighter.

fix the author fields + drop the two severities and i'll pull it in. thanks for these.

PR #251, opened 2026-06-26
feat(modules): add node-red flow api exposure module

TBX3D:
`modules/recon/nodered-flow-exposure.yaml` flags a node-red instance with `adminAuth` unset over its `/flows` admin api, keyed on a `"type":"tab"` entry paired with a `wires` field (so it needs a real flow present), then extracts the first tab label; the open api also accepts a deploy that runs arbitrary javascript and shell through function and exec nodes, and an instance with adminAuth set answers 401.

### DIFF
--- a/internal/modules/nodered_flow_exposure_test.go
+++ b/internal/modules/nodered_flow_exposure_test.go
@@ -1,0 +1,89 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runNodeRedModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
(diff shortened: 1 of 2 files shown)


vmfunc (commented):
solid module. clean and well-scoped.

verified locally: `type: http` so it actually runs through `ExecuteHTTPModule`, the three matchers in `modules/recon/nodered-flow-exposure.yaml` (regex `"type":"tab"`, word `"wires"`, status 200) are AND-combined which is exactly right here since all three are real node-red signatures rather than alternatives, and the `nodered_flow_label` extractor is `type: regex` group 1 so it's an implemented extractor, not a silent kval/json no-op. both patterns are RE2-clean, id is unique. your test suite passes for me, and i like that you pinned the negatives properly: 401 for adminAuth-set, tabs-only-without-wires, wired-graph-without-tab, plain 200, 404 all stay silent. the tab+wires pairing keeps this off catch-all 200 apps, which is the thing i care most about.

no blockers.

two small nits, neither gating:
- consider a `negative` matcher excluding `<html`/`<!DOCTYPE` like `modules/recon/maven-settings-exposure.yaml` does, purely as belt-and-suspenders against an spa that echoes those tokens in a bundle. optional.
- `high` is fair for an unauth flow-config leak. given the /flows deploy path is arbitrary function+exec you could argue `critical`, your call.

pulling it in as-is. thanks for the thorough test coverage.

PR #252, opened 2026-06-26
feat(modules): add caddy and envoy admin exposure modules

TBX3D:
`modules/recon/caddy-admin-exposure.yaml` flags a caddy admin api bound to a routable address over `/config/`, keyed on the `apps` block paired with the `servers` and `handler` fields, then extracts the first handler; the admin api is unauthenticated and meant for loopback, so a 200 exposes the running config and a `POST /load` that replaces it. `modules/recon/envoy-admin-exposure.yaml` flags an exposed envoy proxy admin interface over `/server_info`, keyed on the `hot_restart_version` and `command_line_options` fields paired with the server `state`, then extracts the version; the same interface offers `/config_dump` secrets and `/quitquitquit`.

### DIFF
--- a/internal/modules/server_admin_exposure_test.go
+++ b/internal/modules/server_admin_exposure_test.go
@@ -1,0 +1,101 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runServerAdminModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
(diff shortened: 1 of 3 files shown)


vmfunc (commented):
nice work on these, the two modules are tight and the test file actually exercises them end to end through `ExecuteHTTPModule` with real-hit and near-miss bodies, which is exactly what i want to see.

verified:
- both `type: http`, so they actually run (dns/tcp would be dead)
- `modules/recon/caddy-admin-exposure.yaml`: the three matchers (regex `"apps"\s*:\s*\{`, word `"servers"`+`"handler"` with `condition: and`, status 200) all AND together, and that's correct here since they co-occur in a real `/config/` dump. handler extractor is clean.
- `modules/recon/envoy-admin-exposure.yaml`: same story, `hot_restart_version`+`command_line_options` word matcher, the `"state":"(LIVE|...)"` regex, and status 200 all co-occurring in `/server_info`. version extractor good.
- all four regexes are RE2-clean, no size matchers, extractors are `type: regex` only. ids are unique.
- fp resistance is real: your negatives (apps-only tls block, bare `version`+`state`, plain 200, 404) all stay silent because you keyed on product-specific token pairs, not just status 200. that's the pattern i care about.

optional, take it or leave it:
- severity `high` is fine. you could argue `critical` since both endpoints accept config-replacing POST / `quitquitquit`, but high is acceptable and i won't hold it.

no blockers. pulling this in.

PR #253, opened 2026-06-26
feat(modules): add dagster and mage exposure modules

TBX3D:
`modules/recon/dagster-webserver-exposure.yaml` flags a reachable dagster webserver over its unauthenticated `/server_info`, keyed on the `dagster_webserver_version` and `dagster_version` fields, then extracts the core version; the open instance also serves a graphql api that can launch runs. `modules/recon/mage-status-exposure.yaml` flags a mage data-pipeline server with auth disabled over `/api/status`, keyed on the `statuses`, `scheduler_status` and `repo_path` fields, then extracts the server-side repo path; the open editor can run arbitrary code.

### DIFF
--- a/internal/modules/data_orchestration_exposure_test.go
+++ b/internal/modules/data_orchestration_exposure_test.go
@@ -1,0 +1,99 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runDataOrchModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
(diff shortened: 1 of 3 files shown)


vmfunc (commented):
nice work on these, the FP discipline is exactly right. the word matchers key on quoted json fields unique to each product (`"dagster_webserver_version"`+`"dagster_version"` for dagster, `"statuses"`+`"scheduler_status"`+`"repo_path"` for mage) with `condition: and`, sitting alongside a `status: 200` matcher, so a catch-all app returning 200 html for every path won't trip either one. and you shipped real negatives in `data_orchestration_exposure_test.go` (bare core-version, generic statuses array, plain 200, 404), which is the part most first PRs skip.

checked it end to end:
- both `type: http`, build/vet clean
- matchers combine correctly (word-and AND status, both must pass)
- extractors are `type: regex` (the implemented kind), and `'"dagster_version"\s*:\s*"([^"]+)"'` / `'"repo_path"\s*:\s*"([^"]+)"'` are both RE2-clean, group 1 pulls the value
- ids kebab-case and unique, no dupes in `modules/`

no blockers. couple of small things, take or leave:

- `modules/recon/dagster-webserver-exposure.yaml` description says it "exposes a run-launching graphql api without authentication", but the module only probes `/server_info` (version disclosure). the graphql inference is reasonable since it's the same unauth server, but the module doesn't verify it. maybe soften the wording to what's actually checked.
- `medium` on both is borderline for a version/repo-path leak (leans low), but it's defensible since these endpoints being open implies the whole instance is. your call.
- ignore any note about `author: sif`, that's the repo convention, every existing module uses it.

pulling it in. thanks for the clean tests.

PR #254, opened 2026-06-26
feat(modules): add sidekiq, flower and rq-dashboard exposure modules

TBX3D:
`modules/recon/sidekiq-web-exposure.yaml` flags an unauthenticated sidekiq web dashboard over `/sidekiq/stats`, keyed on the `default_latency`, `redis_version` and `server_utc_time` fields, then extracts the redis version; sidekiq web ships no auth of its own, so a 200 (rather than the app's redirect/401) exposes redis internals and job kill/retry. `modules/recon/celery-flower-exposure.yaml` flags a celery flower instance with its api opened over `/api/workers`, keyed on the `active_queues`, `registered` and `conf` blocks, then extracts the celery version; flower keeps the api disabled unless auth is set or it is explicitly opened, so a 200 means it was deliberately exposed. `modules/recon/rq-dashboard-exposure.yaml` flags an unprotected rq dashboard over `/0/data/queues.json`, keyed on the `queues` array paired with the `failed_job_registry_count` and `scheduled_job_registry_count` fields, then extracts the first queue name.

### DIFF
--- a/internal/modules/job_dashboard_exposure_test.go
+++ b/internal/modules/job_dashboard_exposure_test.go
@@ -1,0 +1,126 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runJobDashModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
+	res, err := modules.ExecuteHTTPModule(context.Background(), srv.URL, def, modules.Options{
+		Timeout: 5 * time.Second,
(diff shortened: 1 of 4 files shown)


vmfunc (commented):
nice work, this is a clean set. all three are `type: http` so they actually run, and the matcher design is exactly right: each one ANDs a `word` matcher (`condition: and` over three product-specific JSON keys) with `status: 200`, so it needs all three unique tokens plus a 200 before it fires. that co-occurrence guard is what keeps these off generic catch-all apps, and your test proves it, the confless/partial/plain-200/404 cases all stay silent. extractors are `type: regex` with clean single-capture RE2 patterns, no size matchers, no kval/json, ids are unique. good instinct including the negative cases in `internal/modules/job_dashboard_exposure_test.go`, that is the part most contributions skip.

one small thing before i pull it in:

- `author: sif` in all three yamls (`celery-flower-exposure.yaml`, `rq-dashboard-exposure.yaml`, `sidekiq-web-exposure.yaml`) should be your handle, not `sif`. convention is the contributor gets the credit.

nits, take or leave:

- `rq-dashboard-exposure.yaml` at medium is defensible, but it exposes redis-backed queue data and allows requeue/delete, so high would be consistent with the other two. your call.
- these lean entirely on the three-token AND for FP protection, which held up empirically. if you ever want belt-and-suspenders you could add a `negative: true` matcher excluding `<html>`/`<!DOCTYPE` like `maven-settings-exposure.yaml` does, but not required here since the bodies are pure json.

fix the `author` field and i will merge it in.

PR #255, opened 2026-06-26
feat(modules): add selenium grid and selenoid exposure modules

TBX3D:
`modules/recon/selenium-grid-exposure.yaml` flags a reachable selenium grid over `/status`, keyed on the `value`/`nodes` envelope paired with the literal `"message":"Selenium Grid ready"`, then extracts the node version; the status read is unauthenticated and the open grid lets anyone drive a browser to internal urls. `modules/recon/selenoid-exposure.yaml` flags a reachable selenoid browser grid over `/status`, keyed on the `total`, `used`, `queued`, `pending` and `browsers` fields, then extracts the first browser image.

### DIFF
--- a/internal/modules/browser_grid_exposure_test.go
+++ b/internal/modules/browser_grid_exposure_test.go
@@ -1,0 +1,115 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runGridModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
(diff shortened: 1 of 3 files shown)


vmfunc (commented):
nice work, this is a clean pair of modules and the test file is genuinely good, the negatives (generic value/nodes blob, browserless capacity, cross-match guard, plain 200, 404) are exactly the cases i'd have asked for.

verified against the engine:
- both `type: http`, so they actually execute (no dead tcp/dns stub trap)
- matchers are combined AND across the list, which is what you want here. `modules/recon/selenium-grid-exposure.yaml` gates on word(`"value"` AND `"nodes"`) + regex(`"message"\s*:\s*"Selenium Grid`) + status 200. the regex is the specificity, generic `value`/`nodes` alone would be FP-prone but that literal message pins it.
- `modules/recon/selenoid-exposure.yaml` leans on the 5-key AND with `"browsers"` doing the real work. total/used/queued/pending are generic on their own but the combined set plus your capacity-without-browsers negative keeps it tight.
- all three regexes compile under RE2, extractors are `type: regex` group 1 (the implemented kind), ids are kebab-case and unique.

two nits, neither blocking:
- `author: sif` in both, could be your handle instead, though every existing module does the same so i won't hold it up.
- severity `medium` is defensible but an open grid is an SSRF + arbitrary-browser-drive surface (internal url reach, file:// reads, plausible RCE via a chrome node), i'd lean `high`. your call, bump it if you agree.

no changes required, pulling this in. thanks for the thorough test coverage.

PR #256, opened 2026-06-26
feat(modules): add prometheus alertmanager exposure module

TBX3D:
`modules/recon/alertmanager-status-exposure.yaml` flags an unauthenticated prometheus alertmanager over `/api/v2/status`, keyed on the `versionInfo`, `config` and `cluster` blocks, then extracts the version; the response carries `config.original`, the full running config, which leaks receiver credentials such as slack, pagerduty and webhook urls and smtp passwords, and an alertmanager behind an auth proxy answers 401.

### DIFF
--- a/internal/modules/alertmanager_exposure_test.go
+++ b/internal/modules/alertmanager_exposure_test.go
@@ -1,0 +1,87 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runAlertmanagerModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


vmfunc (commented):
nice one, this is a clean module. `/api/v2/status` leaking `config.original` with receiver creds is a real high-sev find and you scoped it well.

went through it against the engine and it all holds up:
- `type: http` so it actually executes, matchers combine with AND across the list, so your `word` matcher (`versionInfo` + `config` + `cluster`, `condition: and`) plus `status: 200` means all four signals are required. exactly the FP resistance i want to see. three product-specific json keys, not a bare `status:200 + generic token`.
- extractor is `type: regex` (the only implemented kind), RE2-valid pattern, `group: 1` pulls the version fine. no size matcher, no kval/json, so nothing silently dead.
- id is kebab-case and unique, path uses `{{BaseURL}}`, severity `high` is right for an unauth running-config + credential leak.
- the negative tests in `internal/modules/alertmanager_exposure_test.go` are the good kind: configless, clusterless, plain-200 and 404 all stay silent. i also ran a catch-all 200, a 401, `{}` and a lookalike missing `versionInfo` and it stayed quiet on all of them.

one tiny nit, not a blocker: in `modules/recon/alertmanager-status-exposure.yaml` the `author` should be `TBX3D`, not `sif`, house convention is author = contributor. i'll flip that on the way in so no need to repush unless you want to.

pulling it in. thanks for the thorough test coverage.

PR #257, opened 2026-06-26
feat(modules): add jenkins and nifi exposure modules

TBX3D:
`modules/recon/jenkins-api-exposure.yaml` flags a jenkins controller with anonymous read enabled over `/api/json`, keyed on the root `"_class":"hudson.model.Hudson"` object paired with the `jobs` array, then extracts the first job name; a controller without anonymous read answers 403, so a 200 lists every job, view and node. `modules/recon/nifi-api-exposure.yaml` flags an unsecured apache nifi over `/nifi-api/flow/about`, keyed on the `about` and `version` fields paired with `"title":"NiFi"`, then extracts the version; a secured nifi answers 401, so a 200 grants full code-executing control of the dataflow.

### DIFF
--- a/internal/modules/automation_server_exposure_test.go
+++ b/internal/modules/automation_server_exposure_test.go
@@ -1,0 +1,110 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runAutomationModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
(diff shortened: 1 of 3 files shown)


vmfunc (commented):
nice, this is clean work. two well-scoped modules, both `type: http` so they actually run, and you shipped an end-to-end test with real near-miss cases (`automation_server_exposure_test.go`) which is exactly what i want to see.

the detection logic is right for both:
- `modules/recon/jenkins-api-exposure.yaml` keys on `"_class":"hudson.model.Hudson"` (root-object unique, not a `FreeStyleProject`) AND the `jobs` word AND status 200. that AND across matchers is the correct intent here, and the `hudson.model.Hudson` token is specific enough that a catch-all 200 app won't trip it.
- `modules/recon/nifi-api-exposure.yaml` pairs the `about`/`version` words with the `"title":"NiFi"` regex AND 200. same story, product-unique token so the FP surface is small. the 401 near-miss test covers the secured case.

extractors are `type: regex` (the only implemented kind), all four patterns are RE2-clean (lazy `[^}]*?` and `\s*` only, no lookahead/backrefs), and both ids are unique across `modules/`.

on `author: sif`: ignore any note about switching it to your handle, the whole `modules/` tree uses `author: sif` so you actually matched convention. leave it.

one optional nit, not blocking: nifi is tagged `rce` and an open `/nifi-api` really is full dataflow code-exec, so `critical` would be defensible over `high`. your call, `high` is fine.

no blockers, pulling it in. thanks.

PR #258, opened 2026-06-26
feat(modules): add clickhouse and dgraph exposure modules

TBX3D:
`modules/recon/clickhouse-http-exposure.yaml` flags an unauthenticated clickhouse http interface by running `SELECT version() FORMAT JSON`, keyed on the `meta`/`statistics`/`rows_read`/`bytes_read` json envelope, then extracts the version; the default user has an empty password, so a 200 proves arbitrary sql executes (a password-protected instance answers 403). `modules/recon/dgraph-api-exposure.yaml` flags a reachable dgraph alpha over its unauthenticated `/health`, keyed on `"instance":"alpha"` paired with the `max_assigned` and `lastEcho` fields, then extracts the version; oss alpha has no auth (acl is enterprise) and the open node also serves `/query` and `/admin`.

### DIFF
--- a/internal/modules/database_http_exposure_test.go
+++ b/internal/modules/database_http_exposure_test.go
@@ -1,0 +1,111 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runDBHTTPModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
+	res, err := modules.ExecuteHTTPModule(context.Background(), srv.URL, def, modules.Options{
+		Timeout: 5 * time.Second,
+		Threads: 2,
+	})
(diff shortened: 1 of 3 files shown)


vmfunc (commented):
this is a clean one, pulling it in. thanks TBX3D.

went through both modules and ran them against a pile of hostile responses. the thing i was most worried about, and the thing that kills most clickhouse detections, is that GET / just returns "Ok." which false-positives on half the internet. you sidestepped it exactly right: `clickhouse-http-exposure.yaml` probes `/?query=SELECT version() FORMAT JSON` and keys the word matcher on the json output envelope (`"meta"`, `"statistics"`, `"rows_read"`, `"bytes_read"`, condition: and) plus status 200. i threw a bare 200 "Ok." with no CH headers, a generic `{"status":"ok","meta":"x"}`, an nginx default page, a 401 and a 404 at it. zero findings on all of them, and it fires + extracts the version on a real `SELECT version() FORMAT JSON` body. that's the right shape.

`dgraph-api-exposure.yaml` is solid too: `instance:"alpha"` regex + `max_assigned` + `lastEcho` + 200, so it ignores a zero node and partial health bodies. matchers are AND-combined across the list which is what you want here.

checks that passed: type:http on both (only http actually runs in the engine), matchers are status/word/regex and the extractors are regex, all engine-valid, both regexes are RE2-clean, ids are kebab-case and globally unique. your test file (`database_http_exposure_test.go`) passes 7/7 and i like that it carries real negatives, not just happy-path.

one non-blocking thought, not asking you to change it: dgraph is severity medium while clickhouse is high. defensible as-is since `/health` only proves the alpha is reachable, but keep in mind an exposed dgraph alpha usually means `/query`, `/mutate` and `/alter` are open unauth on that same http interface, which is full read/write/schema-drop. if you ever want to bump it to high i wouldn't argue. good work, merging.

PR #259, opened 2026-06-26
feat(modules): add grafana anonymous access exposure module

TBX3D:
`modules/recon/grafana-anonymous-exposure.yaml` flags a grafana with anonymous access enabled over `/api/search?type=dash-db`, keyed on a `"type":"dash-db"` entry paired with the `uri` and `isStarred` fields, then extracts the dashboard title; default grafana answers 401 here, so a 200 lists every dashboard and the anonymous session can reach backend data sources through the datasource proxy.

### DIFF
--- a/internal/modules/grafana_anonymous_exposure_test.go
+++ b/internal/modules/grafana_anonymous_exposure_test.go
@@ -1,0 +1,93 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runGrafanaModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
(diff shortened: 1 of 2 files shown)


vmfunc (commented):
nice one, this is a clean module. the matcher design is exactly right: `modules/recon/grafana-anonymous-exposure.yaml` gates on the regex `"type"\s*:\s*"dash-db"` AND the word matcher `uri`+`isStarred` (cond: and) AND status 200, so all three have to line up. that token trio is grafana-search-specific, so no firing against a catch-all 200 app, and your test covers the folder-only / no-isStarred / 401 / plain-200 / 404 near-misses. verified it end to end through `ExecuteHTTPModule` and against the engine: matchers are AND-combined across the list, both regexes are RE2-clean, extractor is `type: regex` (implemented) with group 1, and the id is unique. no blockers.

two tiny things before i pull it in, neither is a dealbreaker:

- `info.author: sif` should be your handle. house convention is author = contributor, and you did the work.
- `severity: medium` is defensible, but given anonymous access here reaches backend data sources through the datasource proxy, that's an unauth data-exposure surface and `high` sits better in the rubric. your call, just noting it.

flip the author line (bump severity if you agree) and i'll merge it in. good contribution.

PR #260, opened 2026-06-26
feat(modules): add n8n settings exposure module

TBX3D:
`modules/recon/n8n-settings-exposure.yaml` flags an n8n instance disclosing its configuration over `/rest/settings`, keyed on the `endpointWebhook`, `versionCli` and `instanceId` fields, then extracts the version; the endpoint is served unauthenticated by design so the frontend can load, which leaks the webhook url structure, instance id, configured auth method and whether owner setup is still pending.

### DIFF
--- a/internal/modules/n8n_settings_exposure_test.go
+++ b/internal/modules/n8n_settings_exposure_test.go
@@ -1,0 +1,84 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runN8nModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
+	if err != nil {
+		t.Fatalf("parse %s: %v", file, err)
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


vmfunc (commented):
nice one, this is a clean module. the shape is exactly what i want to see: specific path `{{BaseURL}}/rest/settings`, three product-specific json keys (`endpointWebhook`, `versionCli`, `instanceId`) ANDed together in one `word` matcher, then `status: 200` as a second matcher. since matchers AND across the whole list that reads as "all three n8n keys present AND 200", which is the correct intent, and it's specific enough that it holds up.

pulled the branch and ran it: your `TestN8nSettingsExposureModule` passes, and my own fp fixtures agree, catch-all 200 html, a 401 login blob, and an empty `{}` all come back with 0 findings, the realistic settings json fires exactly 1 and the `n8n_version` extractor pulls the version out. regex is RE2-clean, extractor `type: regex` is the only implemented kind so you're good there, and the id is globally unique.

no blockers.

couple of optional nits, take or leave:
- `severity: medium` leans a touch high for what's a pure info-disclosure (instanceId, auth method, owner-setup state, no creds). for comparison `modules/recon/maven-settings-exposure.yaml` is `high` because it leaks actual passwords. medium is defensible, low/info would also be fine. your call.
- no negative matcher excluding generic `<html>`/`<!DOCTYPE` like the maven module has. empirically you don't need it here, the three ANDed keys already exclude generic pages, but it's cheap insurance if you want to harden against a weird catch-all.

merging this in as is. thanks.

PR #261, opened 2026-06-26
feat(modules): add kafka ui and kafdrop exposure modules

TBX3D:
`modules/recon/kafka-ui-exposure.yaml` flags an exposed kafka ui (provectus/kafbat) console over `/api/clusters`, keyed on the `defaultCluster` and `brokerCount` fields paired with a `"status":"online|offline|initializing"` enum, then extracts the version; the ui ships no auth by default (a login or oauth instance answers 401), so a 200 lists clusters and the session can browse and produce messages. `modules/recon/kafdrop-exposure.yaml` flags an exposed kafdrop ui over `/` requested with an `Accept: application/json` header, keyed on the `summary` block's `preferredReplicaPercent` paired with the `brokers` and `topics` fields, then extracts the first broker host.

### DIFF
--- a/internal/modules/kafka_ui_exposure_test.go
+++ b/internal/modules/kafka_ui_exposure_test.go
@@ -1,0 +1,130 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+const kafkaUIClustersBody = `[{"name":"production","defaultCluster":true,"status":"online",` +
+	`"brokerCount":3,"onlinePartitionCount":128,"topicCount":42,"bytesInPerSec":10485.5,` +
+	`"bytesOutPerSec":20971.0,"readOnly":false,"version":"3.7.0",` +
+	`"features":["TOPIC_DELETION","KAFKA_CONNECT","KSQL_DB"]}]`
+
+const kafdropOverviewBody = `{"summary":{"topicCount":42,"partitionCount":128,` +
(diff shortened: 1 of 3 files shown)


vmfunc (commented):
nice pair of modules, and the test file is genuinely thorough. the near-miss cases (non-enum status, 401, cross-module non-match, missing `preferredReplicaPercent`, plain 200/404) are exactly the shape i want to see, and checking the `Accept: application/json` header actually went out on the wire is a good touch.

no blockers. checked both against the engine:

- both `type: http`, so nothing dead. good, since a raw broker on 9092 could only ever be seen through the ui anyway.
- matcher logic is right. `modules/recon/kafka-ui-exposure.yaml` ANDs three matchers (word `defaultCluster`+`brokerCount`, regex on the `status` enum, status 200) and `modules/recon/kafdrop-exposure.yaml` ANDs the four `summary`/`preferredReplicaPercent`/`brokers`/`topics` words with status 200. specific product tokens, not a generic status:200 + one-word combo, so these won't spray on a catch-all app.
- regex is all RE2-clean, extractor `group: 1` lines up with the single capture group in each. `kafka_version` and `kafka_broker` pull cleanly.
- ids are kebab-case and unique, severity high is right for an unauth console leaking clusters/brokers and letting you produce messages.

one optional thing, not blocking: since kafdrop keys off `/` you could add a `negative` matcher excluding `<html`/`<!DOCTYPE` the way `modules/recon/maven-settings-exposure.yaml` does, as belt-and-suspenders against a root that 200s HTML. the four ANDed json tokens already carry it though, so up to you.

(`author: sif` is fine, that's what the rest of the repo uses.)

pulling this in.

PR #262, opened 2026-06-26
feat(modules): add zookeeper and hadoop namenode exposure modules

TBX3D:
`modules/recon/zookeeper-admin-exposure.yaml` flags an exposed apache zookeeper adminserver over `/commands/monitor`, keyed on the `"command":"monitor"` response envelope paired with the `server_state` field, then extracts the version; the adminserver (default port 8080 since 3.5) has no auth on its read commands, leaking the ensemble role, connection counts and data size. `modules/recon/hadoop-namenode-exposure.yaml` flags an exposed hadoop hdfs namenode jmx interface over `/jmx?qry=Hadoop:service=NameNode,name=NameNodeInfo`, keyed on that bean name paired with the `LiveNodes` and `DeadNodes` attributes, then extracts the software version; the namenode jmx is anonymous unless kerberos spnego is configured, and `LiveNodes` carries every datanode's internal hostname.

### DIFF
--- a/internal/modules/hadoop_zookeeper_exposure_test.go
+++ b/internal/modules/hadoop_zookeeper_exposure_test.go
@@ -1,0 +1,128 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+const zkMonitorBody = `{"command":"monitor","error":null,"version":"3.9.2-a8d7f3e",` +
+	`"server_state":"leader","avg_latency":0,"max_latency":12,"min_latency":0,` +
+	`"packets_received":1532,"packets_sent":1531,"num_alive_connections":3,` +
+	`"outstanding_requests":0,"znode_count":127,"watch_count":4,"ephemerals_count":2,` +
+	`"approximate_data_size":45219,"open_file_descriptor_count":67,"max_file_descriptor_count":1048576}`
+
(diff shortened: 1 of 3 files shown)


vmfunc (commented):
nice work, these two are clean. both are `type: http` hitting the actual web interfaces (namenode `/jmx`, zk adminserver `/commands/monitor`) rather than raw 2181/RPC, so they route through `ExecuteHTTPModule` and actually fire instead of erroring like a tcp/dns stub would.

checked the parts that usually bite module PRs:

- matcher logic is right. each module stacks a specific regex signature (`"command":"monitor"` / the exact `Hadoop:service=NameNode,name=NameNodeInfo` bean) + a word token + `status: 200`, and those all genuinely co-occur in a real response, so AND-across-the-list is the correct intent here, not a case that needed one matcher with `condition: or`.
- fp-resistant. the signature tokens are product-unique, so a catch-all app returning 200 on every path won't trip them. your negative cases (wrong command, 404, 401 kerberos, cross-module body, wrong jmx bean, plain 200 "ok") all stay silent, i re-ran `TestHadoopZooKeeperExposureModules` against main and all 9 subtests pass.
- regexes are RE2-clean, extractors are `type: regex` (the only implemented kind), ids are globally unique, no size/kval/json traps.

on `author: sif`: that's fine, it's what every existing module in `modules/` uses, so you matched convention. no change needed.

only real nit: `severity: medium` is defensible given this leaks version + ensemble/cluster topology + internal datanode hostnames, i'd have accepted low-medium too but medium isn't wrong. leaving it.

merging this in. thanks for the thorough test coverage, the negative + cross-module cases are exactly what i want to see on module PRs.

PR #263, opened 2026-06-26
feat(modules): add eureka and spring boot admin exposure modules

TBX3D:
`modules/recon/eureka-registry-exposure.yaml` flags an exposed netflix eureka registry over `/eureka/apps`, keyed on the `apps__hashcode`, `versions__delta` and `applications` envelope keys (the codec renders the single `_` as `__` in both the xml and json forms), then extracts the first instance ip from either form; eureka ships no auth, so the open registry leaks every instance's internal hostname, ip and ports and accepts rogue registrations. `modules/recon/spring-boot-admin-exposure.yaml` flags an exposed spring boot admin server over `/instances`, keyed on the `registration`, `healthUrl` and `statusInfo` fields, then extracts the first health url; the codecentric server ships no security, so the open registry leaks every app's internal management and health actuator urls (a pivot), and a secured one answers 401 or redirects to login.

### DIFF
--- a/internal/modules/service_registry_exposure_test.go
+++ b/internal/modules/service_registry_exposure_test.go
@@ -1,0 +1,150 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+const eurekaAppsXML = `<applications><versions__delta>1</versions__delta>` +
+	`<apps__hashcode>UP_5_</apps__hashcode><application><name>PAYMENT-SERVICE</name>` +
+	`<instance><instanceId>10.0.0.5:payment-service:8443</instanceId>` +
+	`<hostName>payment-1.svc.internal</hostName><app>PAYMENT-SERVICE</app>` +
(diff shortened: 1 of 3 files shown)


vmfunc (commented):
this is a clean pair of modules, nice work. paths are specific (`/eureka/apps`, `/instances`) and the matchers key on tokens that are actually unique to each product, so the fp surface is basically zero. the double-underscore `apps__hashcode` / `versions__delta` trick for eureka is a good catch, and gating both on `condition: and` + a separate `status: 200` matcher is exactly right given matchers AND across the whole list here. extractors are regex-only with `group: 1`, both the xml and json `ipAddr` forms covered, all patterns RE2-clean. the negative cases in `internal/modules/service_registry_exposure_test.go` (prose-mentions-applications, statusInfo-less body, 401, and the eureka-vs-sba cross-product) are the right things to test.

no blockers. two small things before/when i pull it in:

- `author: sif` in both `modules/recon/eureka-registry-exposure.yaml` and `modules/recon/spring-boot-admin-exposure.yaml` should be your handle (`TBX3D`), credit goes to you not the repo.
- eureka `severity: medium` reads a touch light to me: an open registry leaks the whole internal service map *and* accepts rogue registrations (service impersonation / traffic hijack), which leans high. your call, medium is defensible for pure info-disclosure, just flagging it.

flip the author line and i'll merge it in. the severity bump is optional.

PR #269, opened 2026-07-03
fix(scan): bound crawl breadth with a page budget

TBX3D:
`MaxDepth` bounds recursion depth but nothing bounds breadth, so fetch count grows as branching^depth. Repro before the fix: a depth-3 crawl of one page linking 40 fresh urls issued 1641 real fetches; after, it stops at the 500-page budget and sets `Truncated`.

Verified by `crawl_budget_test.go`: a server linking past the budget caps `visited` at `maxCrawlPages` and flags `Truncated`; an under-budget crawl leaves it false. Race-clean, the counter is atomic since colly fires callbacks across goroutines.

Scope: only the runaway-breadth ceiling. A crawl-wide context deadline is a separate root cause and its own PR.

### DIFF
--- a/internal/scan/crawl.go
+++ b/internal/scan/crawl.go
@@ -17,6 +17,7 @@ import (
 	"net/url"
 	"sort"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/gocolly/colly/v2"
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
counting in OnRequest before the dial is the right spot, abort lands before any fetch. counter is 1..500 pass and 501 aborts, so exactly maxCrawlPages get fetched, clean. Truncated read after Wait() so no race with the callbacks. scoping the context deadline out to its own PR is the right call. in.

PR #270, opened 2026-07-03
fix(modules): guard backup-files against soft-404 html shells

TBX3D:
backup-files was the only recon module with neither a format anchor nor an html guard, so a 200 soft-404 serving the site's html shell tripped the generic tokens (`SECRET_KEY`/`APP_KEY` show up in docs prose) with no dump present.

the guard suppresses a body that opens as an html document, anchored at the body start and covering only `<!doctype` and `<html`, matching the maven-settings/gradle-properties/nuget-config convention.

it deliberately does not cover head/title/body unanchored. a cms dump is full of markup: `wp_posts.post_content` holds rendered html and page builders stash whole documents in serialized postmeta, so a broad negative silently drops the dumps this module exists to catch.

the module shipped without a test; there is now one covering a wordpress dump carrying html in post_content, a plain .env backup, the soft-404 shell and a 404.

### DIFF
--- a/internal/modules/backup_files_exposure_test.go
+++ b/internal/modules/backup_files_exposure_test.go
@@ -1,0 +1,89 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
+:   █▀ █ █▀▀   ·   Blazing-fast pentesting suite                                :
+:   ▄█ █ █▀    ·   BSD 3-Clause License                                         :
+:                                                                               :
+:   (c) 2022-2026 vmfunc, xyzeva,                                               :
+:                 lunchcat alumni & contributors                                :
+:                                                                               :
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+*/
+
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+// runBackupFilesModule runs the backup-files module end to end against a server
+// that returns the same status and body for every path it requests.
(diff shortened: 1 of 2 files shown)


vmfunc (changes requested):
the guard direction is right, backup-files was missing it. two things on the regex though.

it's unanchored, so it matches those tags anywhere in the body, not just at the top. and you added head|title|body to the alternation. a real database.sql / dump.sql from a wordpress or page-builder site has html sitting in post_content and postmeta (<title>, <body>, elementor/divi stash whole <!doctype html> blocks in serialized meta). so this negative trips on the exact dump we most want to catch and the finding goes silent. the comment saying 'real dumps carry no markup' isn't true for cms dumps.

the other three recon guards (maven-settings, gradle-properties, nuget-config) only negate <!DOCTYPE and <html, and they gate behind a specific positive first. backup-files positives are a loose OR of generic tokens, so the broad negative bites harder here.

two ways out: drop it to just <!doctype|html to match the sibling convention, or anchor at the body start (^\s*<(!doctype|html)) so you only suppress responses that actually open as an html document, which is what the soft-404 shell does. either kills the docs-prose FP without eating wp dumps.

TBX3D:
both points are right and the wp case is the one that mattered. anchored it and
dropped back to the two document-opening tags, matching the
maven-settings/gradle-properties/nuget-config convention:

  '(?i)^\s*<(!doctype|html)[ >]'

the module had no test at all, which is why the FN was invisible. there is one
now covering a wordpress dump whose post_content holds <title>/<body> and whose
postmeta stashes a serialized <!doctype html> blob, a plain .env backup, the
soft-404 shell and a 404. the dump case fails against the old regex.

also correcting my own commit message: "real dumps carry no markup" was just
wrong for cms dumps.

TBX3D:
addressed: the guard is now anchored at body start and narrowed to the two document-opening tags — `(?i)^\s*<(!doctype|html)[ >]`. dropped `head|title|body` from the alternation, since a real cms dump carries markup inside `wp_posts.post_content` and serialized postmeta, and an unanchored guard covering those tags silently drops the exact files this module exists to catch.

added a regression test (`TestBackupFilesExposureModule`) with a wordpress dump that embeds `<!doctype html><html>` inside `_elementor_data` postmeta: it's flagged, while an html soft-404 shell that opens with `<!DOCTYPE` is suppressed. rebased on current main.

PR #271, opened 2026-07-03
feat(modules): add active web-vuln probe modules

TBX3D:
Five active-probe modules for `modules/http`: `ssti-reflected` (arithmetic-eval proof, matches 1787569 = eval of 1337*1337), `command-injection` (matches id output shape `uid=N(...) gid=N(...)`, not the payload echo), and error-based `xpath`, `nosql`, `ldap`.

Each keys on a transformed response the payload cannot produce by reflection, so a benign echo of the input cannot trigger a finding. Verified with an eval-server vs echo-server pair: eval fires, echo stays silent for every module.

### DIFF
--- a/modules/http/command-injection.yaml
+++ b/modules/http/command-injection.yaml
@@ -1,5 +1,3 @@
-# OS Command Injection Detection Module
-
 id: command-injection
 info:
   name: OS Command Injection
--- a/modules/http/command-injection.yaml
+++ b/modules/http/command-injection.yaml
@@ -1,0 +1,53 @@
+# OS Command Injection Detection Module
+
+id: command-injection
+info:
+  name: OS Command Injection
+  author: sif
+  severity: critical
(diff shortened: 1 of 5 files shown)


vmfunc (approved):
consistent with the existing http modules and the executor already handles all of this (paths x payloads, word/regex, or-condition, extractors). the two eval-based ones are the strong ones: command-injection keys on the real `uid=N(...) gid=N(...)` shape and ssti on the 1787569 product, so a plain echo of the payload can't trip either. good.

one nit, doesn't block: in ldap-injection-error the `Protocol error occurred` matcher is weak on both ends. openldap emits "Protocol error", not "...occurred", so it probably misses the real signal, and it's generic enough to false-positive on any unrelated protocol error, and since the list is or'd it taints the whole module. drop it or tighten it, the rest carry the module fine.

approving.

PR #272, opened 2026-07-03
feat(modules): add cockpit and gitea fingerprint modules

TBX3D:
two info-severity fingerprint modules.

`cockpit-panel` detects an exposed Cockpit web console: keys on the login page's three stable dom anchors (the user and password input ids and the cockpit login script path) under an and-condition, so a generic 200 cannot trip it.

`favicon-gitea` matches Gitea's default favicon. gitea reskinned that icon twice, so one hash never covered the deployed population. hashing `public/assets/img/favicon.png` at each release tag with `FaviconHash` gives:

```
975625564    8268 bytes  v1.4-v1.8
1969970750   5576 bytes  v1.9-v1.12
-564935264   4873 bytes  v1.13
-754147112   4351 bytes  v1.15-v1.24
-1668137428  4335 bytes  v1.25 onward
```

-1668137428 is confirmed live against gitea.com (1.27.0+dev) and demo.gitea.com (1.28.0+dev), both serving /favicon.ico as that png. -754147112 spans nine minor releases and was the real coverage gap, so both are in the canonical table and the module lists both. pre-v1.15 icons stay out, those releases are long EOL.

the pre-existing -1893514588 entry is kept. it matches none of the tagged release icons above so its provenance is unclear, but removing coverage that might be real is worse than carrying an unverified entry. codeberg (forgejo) hashes 1315211563, so none of these collide with it.


vmfunc (changes requested):
cockpit-panel is right, status 200 anded with the three dom anchors, generic 200 won't trip it. that half's in.

the gitea half has two problems.

favicon-gitea.yaml is dead. the yaml module executor has no favicon matcher: module.go Matcher has no hash field and executor.go checkMatcher default-returns-false for anything that isn't status/word/regex. matchers are and'd, so that template is status 200 && false == never fires. favicon fingerprinting doesn't go through the module path at all, it lives in internal/scan/favicon.go behind -favicon. so this file matches nothing. drop it.

and favicon.go line 70 isn't "no drift", it rewrites the existing gitea entry from -1893514588 to -1668137428. if the old hash was a valid older-gitea icon you just lost coverage for it. add the new one as a second entry mapping to gitea instead of replacing, and tell me where -1668137428 came from... which gitea version, verified how. that map is the only thing that actually detects gitea here.

cockpit stays, rework the gitea side and this is in.

TBX3D:
the favicon.go half is entirely fair and I have fixed it, see below. on the
yaml half I got a different result, so flagging it rather than just deleting
the file.

the yaml executor does have a favicon matcher on main: Matcher carries
`Hash []int64 \`yaml:"hash,omitempty"\`` (module.go:95) and checkMatcher has
`case "favicon": return checkFaviconHash(body, m.Hash)` (executor.go:391). I ran
the module end to end against the real bytes from gitea.com/favicon.ico and it
fires. there is also TestFaviconDemoModulesMatchCanonicalMap already gating
modules/info/favicon-*.yaml against the canonical table. so I have kept the file.

on the hash: you are right that it was a replacement and not "no drift", and my
commit message claiming otherwise was wrong. to answer where -1668137428 came
from, I hashed public/assets/img/favicon.png at each release tag with
fingerprint.FaviconHash:

  975625564    8268 bytes  v1.4-v1.8
  1969970750   5576 bytes  v1.9-v1.12
  -564935264   4873 bytes  v1.13
  -754147112   4351 bytes  v1.15-v1.24
  -1668137428  4335 bytes  v1.25 onward

-1668137428 is also confirmed live against gitea.com (1.27.0+dev) and
demo.gitea.com (1.28.0+dev), both serving /favicon.ico as that 4335-byte png.

that turned up a bigger gap than the one under review: -754147112 covers nine
minor releases and neither the old nor the new entry had it. so the table now
carries both, and the module lists both.

I kept -1893514588 rather than dropping it. it matches none of the tagged release
icons, so I cannot tell you what it is, but removing coverage that might be real
is the thing you were objecting to. flagging it in case you know where it came
from, since you would know better than I would.
codeberg (forgejo) hashes 1315211563, so nothing here collides with it.

the table also moved to internal/fingerprint after #358, so the change is there
now, and the count pin in TestLookupFaviconTech went 10 -> 12.

TBX3D:
pushed a revision that addresses the favicon points:

- the module engine now has a `favicon` matcher (`internal/modules/executor.go` `case "favicon"` -> `checkFaviconHash`), so `favicon-gitea.yaml` is live rather than dead config. it probes both `/assets/img/favicon.png` and the legacy `/img/favicon.png` and or's the default-icon hashes.
- provenance for the hashes, since you asked where they came from: i ran `FaviconHash` over `favicon.png` checked out at each tagged release. `-754147112` is v1.14 through v1.24, `-1668137428` is v1.25 onward. the bytes are identical whether gitea serves them from `public/img/` (pre-v1.21) or `public/assets/img/` (after), so the hash is path-independent. i corrected an earlier comment of mine that had guessed a pre-v1.15 range that the bytes did not actually support. the pre-existing `-1893514588` entry is left untouched.
- added `TestFaviconGiteaModule`, which serves the real v1.24 and v1.25 icons (checked in under `internal/modules/testdata`) and asserts each fires on either path while an unrelated icon does not, plus `TestGiteaFaviconGenerationsAllResolve` on the table itself.

gofmt/build/vet/golangci-lint/test all green.

PR #273, opened 2026-07-03
feat(modules): add disable-redirects http option

TBX3D:
A module that fingerprints a redirect itself (open-redirect proofs, a `Location` header, a 3xx status) could not see it: the executor followed 3xx to the final response before matchers ran, with no opt-out.

Add a `disable-redirects` http field that scopes an `ErrUseLastResponse` policy to a shallow copy of the module's client, so the shared httpx transport keeps following redirects for every other module in the run. Test drives both paths against a 302-to-landing server: followed resolves to the 200 landing (302 matcher silent), disabled stops at the 302 (matcher fires).

### DIFF
--- a/internal/modules/executor.go
+++ b/internal/modules/executor.go
@@ -72,6 +72,18 @@ func ExecuteHTTPModule(ctx context.Context, target string, def *YAMLModule, opts
 		}
 	}
 
+	// disable-redirects only applies to this module's requests; opts.Client may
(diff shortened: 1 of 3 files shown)


vmfunc (approved):
shallow-copy of the client to scope CheckRedirect is exactly right, mutating the shared httpx instance would've broken redirect following for every other module in the run. body still gets closed, matchers see the 302. test covers both paths.

in.

PR #274, opened 2026-07-03
feat(scan): add tls certificate recon module

TBX3D:
A `-tls-cert` scan that actively dials the target's tls port and mines the leaf certificate: SAN entries become candidate subdomains, issuer and validity feed posture flags (self-signed, expired, expiring soon, wildcard).

Unlike the existing passive crt.sh and certspotter feeds this probes the target directly, so it catches certs never logged to a public ct log (short-lived or internal-CA certs) at the cost of touching the target. `InsecureSkipVerify` is deliberate: the module mines whatever cert is presented, self-signed or expired included, rather than only certs that validate.

Self-signed detection compares raw issuer/subject der and verifies the cert against its own key, not `CheckSignatureFrom`, which false-negatives on leaf certs that omit CA key-usage bits. Tests generate self-signed leaves in-process to exercise the wildcard, subdomain, and expiry paths.


vmfunc (approved):
self-signed check is the right call. RawIssuer==RawSubject plus CheckSignature against the cert's own key sidesteps the CheckSignatureFrom ca-keyusage false-negative, which is exactly the trap on internal-CA leaves. wildcard detection off raw DNSNames before normalizeHost strips the *. is correct, and normalizeHost keeps the target out of new_subdomains while Wildcard carries the signal separately. conn closed, dialer timeout set, InsecureSkipVerify documented and scoped to recon. tests cover the wildcard/subdomain/expiry paths.

only thing: it's CONFLICTING with main and the new file imports github.com/vmfunc/sif/... while main is on github.com/dropalldatabases/sif. rebase and fix the import path to match and it's in.

nitpick, not blocking: SANs only reads DNSNames, so CN-only or IP-SAN certs get skipped. fine for subdomain mining, worth a follow-up if you want ip coverage.

PR #275, opened 2026-07-03
feat(modules): add zabbix and icinga monitoring exposure modules

TBX3D:
zabbix-panel and icinga-panel fingerprint the monitoring frontends by two framework-specific login-page classes under an and-condition, so a generic 200 cannot trip them. strings were verified against current upstream login markup.

zabbix-api-exposure posts the unauthenticated `apiinfo.version` json-rpc call and matches on status 200 plus a jsonrpc-2.0 dotted-result body, which nothing but apiinfo.version answers with. it does not match on the response content-type: `application/json-rpc` is a request precondition, `ui/api_jsonrpc.php` accepts application/json-rpc, application/json or application/jsonrequest and answers 412 otherwise, then always sets `header('Content-Type: application/json')`. since matchers are and'd, requiring json-rpc back meant the module never fired against a live instance.

the `low` severity matches the compatible-shape (not product-string) nature of that evidence. the test drives the real application/json response, keeps application/json-rpc as a second firing case for proxies that echo the request type, and pins a json-rpc error body as silent.

### DIFF
--- a/internal/modules/zabbix_api_exposure_test.go
+++ b/internal/modules/zabbix_api_exposure_test.go
@@ -1,3 +1,15 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
+:   █▀ █ █▀▀   ·   Blazing-fast pentesting suite                                :
+:   ▄█ █ █▀    ·   BSD 3-Clause License                                         :
+:                                                                               :
+:   (c) 2022-2026 vmfunc, xyzeva,                                               :
+:                 lunchcat alumni & contributors                                :
+:                                                                               :
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+*/
+
 package modules_test
 
 import (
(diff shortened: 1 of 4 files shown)


vmfunc (changes requested):
panels are good, the two-class and-condition on each is specific enough that a generic 200 won't trip them.

the api module i want to hold on. line 27, the header matcher requires the response content-type to contain `application/json-rpc`. pretty sure that's the request content-type, not the response... real zabbix api_jsonrpc.php answers apiinfo.version with `Content-Type: application/json`. if that's right, since matchers are and'd this never fires against a live instance and the module is dead. the test doesn't catch it because it hardcodes application/json-rpc on the mock, so it's just asserting the module's own assumption back at itself.

verify against a real zabbix (any 6.x/7.x). if it comes back application/json like i think, either drop the header matcher entirely (status 200 + the jsonrpc-2.0/dotted-result regex is already specific to apiinfo.version, nothing else answers that shape) or make it condition: or with both types. fix that and it's in.

TBX3D:
you were right and it is worse than a mismatch, the module could never have
fired. checked ui/api_jsonrpc.php upstream:

  $allowed_content = [
      'application/json-rpc'    => 'json-rpc',
      'application/json'        => 'json-rpc',
      'application/jsonrequest' => 'json-rpc'
  ];
  ...
  header('Content-Type: application/json');

so those three types are a request precondition (412 otherwise) and the response
is always application/json. dropped the response header matcher entirely, as you
suggested: status 200 plus the jsonrpc-2.0 dotted-result regex is already
specific to apiinfo.version.

the test was asserting the bug back at itself, it served the mock as
application/json-rpc and required the real application/json case to produce zero
findings. it now drives the real shape, keeps json-rpc as a second firing case
for proxies that echo the request type, and pins a json-rpc error body as silent.

TBX3D:
one more on the api module: the body regex required "jsonrpc":"2.0" immediately
before "result", which assumes json object key order. that holds for zabbix
today but is not guaranteed by anything, so a reordered body like
{"id":1,"jsonrpc":"2.0","result":"7.0.5"} would have been missed.

split into two and'd patterns so the match is order-insensitive, with a test
case for the reordered shape.

TBX3D:
same way:

TBX3D:
you're right — that was the request content-type leaking into the response matcher. dropped the response content-type matcher entirely: `Content-Type: application/json-rpc` now stays only as the *request* precondition (zabbix 412s without one of json-rpc/json/jsonrequest on the request), and the module keys on status 200 + two and'd regexes (`jsonrpc:2.0` and a dotted `result` version), which is specific to `apiinfo.version`.

fixed the test to match real behavior: the mock now answers `Content-Type: application/json` (real `api_jsonrpc.php`) and the module still fires; added cases that it also fires on `application/json-rpc`, and does *not* fire on a non-version `"result":"pong"` or on a jsonrpc error object. panels untouched.

PR #276, opened 2026-07-03
fix(frameworks): stop prose mentions from faking strapi and cakephp

TBX3D:
The Strapi and CakePHP detectors keyed on a bare body word (`strapi`, `cakephp`). `MatchSignatures` normalizes matched weight by total signature weight, so a lone body signature always scored a full 1.0, and the single-argmax reducer then let one prose mention on an unrelated page outrank and suppress the site's real, multi-marker framework (a WordPress page with `wp-content`/`wp-includes` lost to a listicle that merely named Strapi).

Key on the structural markers instead: the `X-Powered-By` poweredBy header for Strapi and the `CAKEPHP` session cookie for CakePHP, both header-only. Tests fail without the fix (prose suppresses the real framework) and pass with it. CakePHP is folded in as the same root-cause bug class rather than a separate PR.

### DIFF
--- a/internal/scan/frameworks/detect_test.go
+++ b/internal/scan/frameworks/detect_test.go
@@ -1004,6 +1004,57 @@ func TestDetectFramework_Strapi(t *testing.T) {
 	}
 }
 
+// a page that merely names the framework in prose (a listicle, a comparison)
+// must not be fingerprinted as that framework, as the old bare body words did.
+func TestDetectFramework_ProseMentionsAreNotDetections(t *testing.T) {
+	cases := map[string]string{
+		"Strapi":  `<h1>Is Strapi Right For You?</h1><p>we compare strapi with other headless cms platforms.</p>`,
+		"CakePHP": `<h1>Choosing a PHP framework</h1><p>cakephp, laravel and symfony all have strong communities.</p>`,
+	}
+	for name, body := range cases {
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(`<!DOCTYPE html><html><body>` + body + `</body></html>`))
+		}))
+		result, err := frameworks.DetectFramework(server.URL, 5*time.Second, "")
+		server.Close()
+		if err != nil {
+			t.Fatalf("%s: unexpected error: %v", name, err)
+		}
+		if result != nil && result.Name == name {
+			t.Errorf("false positive: prose mentioning %s detected as %s (%.2f)", name, name, result.Confidence)
+		}
+	}
+}
+
+// the bare body words let a single prose mention outscore the site's real,
+// multi-marker framework: MatchSignatures normalizes by total signature weight,
+// so a lone signature always resolved to a full 1.0, and the single-argmax
+// reducer then swapped the true framework for the spurious one.
+func TestDetectFramework_ProseDoesNotSuppressRealFramework(t *testing.T) {
+	for _, mention := range []string{"strapi", "cakephp"} {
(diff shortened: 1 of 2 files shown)


vmfunc (changes requested):
direction is right, header-only structural markers beat bare body words for both of these.

but this is stacked and against a stale base, the diff doesn't match main:

- cakephp: the `CAKEPHP` header signature you're keying on comes from #133 (d34db55), which isn't on main. main's cakephp is still {cakephp 0.4, cake 0.2}, no header marker. so this hunk depends on #133 landing first, or fold it into #133.

- strapi: your base predates the `/api/` signature that landed in 29d94e5. main's strapi is {strapi 0.4, /api/ 0.2}. as this merges you end up with {Strapi header 0.4, /api/ 0.2}, and /api/ on its own scores 0.2/0.6 -> sigmoid 0.58, over the 0.5 floor. so any page with "/api/" in the body still resolves to strapi. that's the same FP class you're closing, your prose test just doesn't include "/api/" so it stays green. drop /api/ while you're in there.

rebase on current main and reconcile with #133 + the sigmoid recenter, then this is good. minor: strapi's poweredBy header is often disabled in prod, so it's a FN tradeoff on hardened installs, fine by me but noting it.

TBX3D:
good call on the stale base, that was the real problem here. once rebased, the
reconciliation you were asking for turns out to have already happened upstream.

cakephp: the CAKEPHP header signature is on main now, so there is no dependency
on #133 left. after the rebase cakephp is {CAKEPHP 0.4, HeaderOnly}, the bare
body word gone.

strapi: main's strapi is a single {strapi 0.4} with no /api/ signature, so the
0.2/0.6 -> 0.58 path you were worried about does not exist to close. after the
rebase it is {Strapi 0.4, HeaderOnly} on X-Powered-By alone. I grepped the whole
detectors tree for an "/api/" signature and there is none.

so the diff is now just the two header-only markers against current main, no
stacking. noted on the poweredBy FN tradeoff for hardened strapi installs, that
still stands and I think it is the right trade.

TBX3D:
one more on this branch, from the same root cause as #287.

moving strapi and cakephp to header-only markers closed the prose false positive
but did not fully close the class, because neither signature was scoped to a
header:

  {Pattern: "Strapi", Weight: 0.4, HeaderOnly: true}

a HeaderOnly signature with no Header runs through containsHeader, which
substring-matches every header name and every header value. so the word still
fired from any header that merely named it:

  X-Cache-Tags: cms,Strapi,blog                    -> "Strapi"  0.9991
  Access-Control-Expose-Headers: CAKEPHP, X-Total  -> "CakePHP" 0.9991

both now carry the header they actually live on, Header: "X-Powered-By" and
Header: "Set-Cookie", and there is a regression test for the two cases above.
the existing prose and true-positive tests still pass.

my comment in the previous version claimed the detector keyed on X-Powered-By
when the code did not, which is the sort of thing you flagged on #287. fixed
both.

TBX3D:
same way:

TBX3D:
rebased on current main and reduced to the change that's still needed. two of the review points resolved themselves on main in the meantime:

- strapi is already scoped to `X-Powered-By` on main and the bare `/api/` body word is gone, so there's nothing left to drop there.
- cakephp's `CAKEPHP` marker is on main independent of #133, so this no longer depends on that PR.

the remaining delta is scoping cakephp's `CAKEPHP` from an unscoped `HeaderOnly` marker to `Set-Cookie`. unscoped, a `HeaderOnly` signature runs through `containsHeader` and matches header *values* too, so an `Access-Control-Expose-Headers: CAKEPHP` list resolved as CakePHP at ~0.999 — the same fp class, moved not closed. added regression tests for both frameworks covering prose mentions, unrelated-header values, and the argmax-suppression case (a prose mention no longer outranks a real WordPress fingerprint).

PR #277, opened 2026-07-03
feat(modules): add host-header injection and xxe probe modules

TBX3D:
Two active-probe modules for `modules/http`.

`host-header-injection` sets an rfc 2606 `.invalid` marker in the forwarded-host headers and fires only when it comes back inside an absolute-url construct (`href`/`src`/`action`/`content`/`url`), so a benign echo into visible text or an attribute cannot trigger it.

`xxe-error-based` expands a `SYSTEM` entity pointing at a local file and matches only that file's contents, so an echo of the payload (a path string) cannot satisfy it. This reuses the same matcher discipline as the existing `lfi-basic` module.

### DIFF
--- a/modules/http/host-header-injection.yaml
+++ b/modules/http/host-header-injection.yaml
@@ -1,4 +1,3 @@
-# Host Header Injection Detection Module
 id: host-header-injection
 info:
   name: Host Header Injection
--- a/modules/http/host-header-injection.yaml
+++ b/modules/http/host-header-injection.yaml
@@ -1,0 +1,48 @@
+# Host Header Injection Detection Module
+id: host-header-injection
+info:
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
host-header one is right. marker's rfc-2606 unique and gating on the href/src/action/content/url anchor means a plain text echo can't trip it, only reflection into a real link construct does. confirmed the regex behaves that way. xxe matcher keys on the target file's contents not the path, same discipline as lfi-basic, good.

one thing, not a blocker: xxe-error-based is actually in-band reflective xxe, error-based is the param-entity + forced-parse-error oob trick. name's misleading for reports, rename to xxe-inband or xxe-reflective when you get a sec. in either way.

PR #278, opened 2026-07-03
feat(modules): add mongo-express, redis-commander and pgweb exposure

TBX3D:
Detect unauthenticated web database browsers (Mongo Express, Redis Commander, pgweb), which ship with no default auth and hand over a full data browser once reachable, a distinct risk class from the login-gated phpMyAdmin/Adminer panels already covered.

Each keys on an app-specific anchor (the `ME_SETTINGS` global, the `redisCommander.js` script, `ace-pgsql.js`) under an and-condition so a branded soft-404 cannot trip it. mongo-express additionally requires the database-list heading: its login form renders the same settings global, title and logo but serves no data, so keying on those alone would false-positive on a data-less form. Anchor strings verified against current upstream markup.

### DIFF
--- a/internal/modules/db_browser_panel_exposure_test.go
+++ b/internal/modules/db_browser_panel_exposure_test.go
@@ -1,0 +1,213 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runDBBrowserModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
+	def, err := modules.ParseYAMLModule(file)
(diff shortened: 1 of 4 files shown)


vmfunc (approved):
the mongo-express login-vs-datalist split is the right call, keying on globalThis.ME_SETTINGS alone would trip on the connection form since login.html extends the same layout. requiring >Databases</h4> pins it to an actual rendered listing. redis/pgweb both anchor on the specific script path and AND a panel marker, so no generic-200 trips. matchers and the version extractor are all supported types, negative tests cover the soft-404 and non-200 cases.

one thing, not blocking: these are tagged exposure/recon but they land in modules/info/ while backup-files and git-exposed sit in modules/recon/. might want them in recon/ for consistency, your call.

in.

PR #279, opened 2026-07-03
feat(modules): add laravel telescope dashboard exposure module

TBX3D:
Detect an exposed Laravel Telescope dashboard, which reveals requests, queries, exceptions and dumped variables.

Telescope's middleware aborts with 403 before render when gated, so a 200 carrying both the Telescope title and the Vue mount id (`id="telescope" v-cloak`) proves the gate is open rather than merely hinting at the route. The and-condition on the mount id keeps an unrelated page that merely names telescope, or a soft-404, from matching. Both tokens are plain markup outside Blade's `{{ }}` interpolation, so they render byte-for-byte and are stable across Telescope major versions.

### DIFF
--- a/internal/modules/telescope_exposure_test.go
+++ b/internal/modules/telescope_exposure_test.go
@@ -1,0 +1,53 @@
+package modules_test
+
+import "testing"
+
+// laravel-telescope-exposure fires only when the dashboard actually renders: a
+// 200 carrying both the Telescope title and the Vue mount id. A 403-gated
+// install, a soft-404 that reflects the word, and an unrelated "telescope" page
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
checked both tokens against telescope's layout.blade.php. `<title>Telescope{{ config('app.name') ... }}</title>` and `<div id="telescope" v-cloak>` render exactly to the two words, and the `- ([^<]+)` extractor lands on the app_name form. tight enough that the astronomy `id="telescope-list"` page stays silent even without v-cloak, so the mount id is doing real work, not padding. in.

PR #281, opened 2026-07-03
fix(scan): correlate cname before confirming subdomain takeover

TBX3D:
the live-response path flagged any host whose body matched a provider
fingerprint without checking dns, so a page merely containing a
provider 404 string was reported as a takeover. resolve the cname and
only confirm when it maps to the same provider; a contradicting cname
kills the flag. providers with a body signature but no apex to
correlate, and unresolvable lookups, degrade to a medium potential
finding via a new confidence field rather than confirming or dropping.

### DIFF
--- a/internal/finding/finding.go
+++ b/internal/finding/finding.go
@@ -577,10 +577,17 @@ func flattenTakeover(target string, rs []scan.SubdomainTakeoverResult) []Finding
 		if !t.Vulnerable {
 			continue
 		}
+		// a "potential" result only matched a body fingerprint and could not be
+		// checked against the cname (lookup unavailable), so it does not earn
+		// the same severity as a cname-confirmed takeover.
+		sev := sevTakeover
+		if t.Confidence == "potential" {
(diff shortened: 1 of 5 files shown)


vmfunc (approved):
logic's right, and the three-way split (confirmed / dropped-on-contradiction / potential) is the correct shape for this. reusing danglingProvider for the body path so it lines up with the no-such-host path is the right call.

two non-blocking things:

- "confirmed"/"potential" are bare literals in scan, and finding.go compares against a bare "potential". give them named consts off SubdomainTakeoverResult so the two packages can't drift on a typo.

- serviceCorrelatable leans on danglingProvider's apex suffixes, which are region-blind. a real regional s3 endpoint (bucket.s3.us-west-2.amazonaws.com) won't match .s3.amazonaws.com, so for a correlatable provider that path drops the finding entirely instead of degrading to potential. it's the same narrowness the dangling path already has so nothing new here, but worth a follow-up on the apex map.

in.

PR #282, opened 2026-07-03
fix(scan): require a real bucket listing body before flagging s3 exposure

TBX3D:


vmfunc (approved):
good catch on the 200-alone false positive, and the tests actually pin the pre-fix bug (200 + AccessDenied body). LimitReader keeps the read bounded and DrainClose still returns the conn, so no leak.

one thing worth noting: the denial markers are matched with strings.Contains over the whole body, not scoped to <Code>...</Code>. so a genuinely listable bucket whose listing contains an object key like AccessDenied.txt or a PermanentRedirect/ prefix would get scored not-public. edge case, but if you want to tighten it later, matching <Code>AccessDenied</Code> would remove the collision. not blocking.

in.

PR #283, opened 2026-07-03
fix(scan): report the found url in dork results and dedupe by url

TBX3D:


vmfunc (approved):
good catch on the actual bug here, Url was holding the dork query the whole time so key("dork", d.Url) in flattenDork was keying findings on the query instead of the hit. groupDorkResults dedup is correct, empty-url drop and order-preserving map both right. in.

PR #284, opened 2026-07-03
fix(recon): tighten directory-listing matchers to real listing furniture

TBX3D:
require actual autoindex markup (parent-directory anchors, sort links,
bracketed parent link) instead of loose title text, so prose pages
titled "Index of /..." and iis tutorials no longer false-positive.
apache non-fancy parent-directory anchor tolerates surrounding
whitespace to avoid a false negative.

### DIFF
--- a/internal/modules/directory_listing_exposure_test.go
+++ b/internal/modules/directory_listing_exposure_test.go
@@ -1,0 +1,240 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+func runDirectoryListingModule(t *testing.T, file string, status int, body string) *modules.Result {
+	t.Helper()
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
furniture-anchoring instead of loose title text is the right call, this kills the 'index of /...' prose FP class cleanly. the subtle one is alt2's `\s*parent directory\s*` swallowing the non-fancy `"> ` leading space, and you pinned it with the plain-mod_autoindex test so it won't silently regress. RE2 so the `.*?` bridges are linear, no ReDoS.

two things, neither blocking:
- alt1 keys on the `../` self-anchor. nginx autoindex doesn't emit that at the filesystem root, so a listing served at `/` itself slips past that alternative. the subdir probes still catch it, fine to leave.
- the caddy alt is welded to the stock browse template ("folder path" breadcrumb + `class='listing'`). a customized template misses. inherent to the heuristic.

reads as the tightening follow-up to the looser version, on current main the file's net-new so it applies clean and stands on its own. in.

PR #285, opened 2026-07-03
fix(frameworks): stop gatsby and svelte prose from faking detection

TBX3D:


vmfunc (approved):
clean fix. the math checks out: after dropping the prefixes, gatsby needs ___gatsby or /page-data/ and svelte needs __svelte or svelte/internal, all of which only show up in a real bundle, and each still clears the 0.3 center on a single strong hit so recall on genuine pages holds. no false-negative regression that matters. /page-data/ over page-data.json is the better call too, its the actual served path structure, not a string prose can drop. in.

PR #286, opened 2026-07-03
fix(modules): guard ignition, web.config and appsettings against soft-404 shells

TBX3D:


vmfunc (approved):
good change. the ignition one is the real win: dropping the bare `can_execute_commands` word matcher for `"can_execute_commands"\s*:\s*(?:true|false)` anchors on the json shape the endpoint actually emits, so an spa bundle referencing the field as a js property stops tripping it. negative-matcher AND semantics check out in checkMatcher, and the guard regexes are fine under re2.

one thing to keep in mind, not blocking: the `<(!doctype|html|head|title|body)[ >]` guard will also suppress a genuine web.config that embeds an html sample inside an xml comment. narrow, and worth it against the docs-page fp, just noting it.

this sits on top of the grafana/debug+webserver module bundle and the app-config PR, not main. fine to land once those are in.

in.

PR #287, opened 2026-07-03
feat(scan): add cdn provider detection pool

TBX3D:
detect cdn providers via a dedicated detection pool separate from the
framework single-argmax reducer, so a cdn match cannot suppress a
framework match. new cdn-detection module fetches once, runs
frameworks.DetectCDN, and emits one info finding for the highest-confidence
provider (DetectCDN is itself single-argmax).

the brand-word signatures are scoped with `Header: Server` / `Header: Via`, which MatchSignatures routes through headerValueContains rather than the all-headers containsHeader. a test pins the case that protects: an nginx origin whose CSP and Link headers reference cdnjs.cloudflare.com, a cloudfront bucket, a vercel.app frame-ancestor and a netlify api, with no edge header anywhere, stays unreported.

### DIFF
--- a/internal/scan/builtin/cdn_module.go
+++ b/internal/scan/builtin/cdn_module.go
@@ -1,0 +1,91 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
+:   █▀ █ █▀▀   ·   Blazing-fast pentesting suite                                :
+:   ▄█ █ █▀    ·   BSD 3-Clause License                                         :
+:                                                                               :
+:   (c) 2022-2026 vmfunc, xyzeva,                                               :
+:                 lunchcat alumni & contributors                                :
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+*/
+
+package builtin
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/vmfunc/sif/internal/httpx"
+	"github.com/vmfunc/sif/internal/modules"
+	"github.com/vmfunc/sif/internal/scan/frameworks"
+)
+
+// cdnMaxBodySize caps the response body read, mirroring frameworks.maxBodySize.
+const cdnMaxBodySize = 5 * 1024 * 1024
+
+type CDNModule struct{}
+
+func (m *CDNModule) Info() modules.Info {
+	return modules.Info{
+		ID:          "cdn-detection",
+		Name:        "CDN/Hosting Provider Detection",
+		Author:      "sif",
+		Severity:    "info",
+		Description: "Fingerprints the cdn/edge/hosting provider fronting a target from response headers",
+		Tags:        []string{"recon", "cdn", "hosting", "fingerprint"},
+	}
+}
+
+func (m *CDNModule) Type() modules.ModuleType {
+	return modules.TypeHTTP
+}
+
+// Execute fetches the target and runs the CDN detector pool over the response,
+// independent of framework detection (see cdnRegistry in
+// internal/scan/frameworks/cdn.go).
+func (m *CDNModule) Execute(ctx context.Context, target string, opts modules.Options) (*modules.Result, error) {
+	client := opts.Client
+	if client == nil {
+		client = httpx.Client(opts.Timeout)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, target, http.NoBody)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(io.LimitReader(resp.Body, cdnMaxBodySize))
+	if err != nil {
+		return nil, err
+	}
+
+	result := &modules.Result{
+		ModuleID: m.Info().ID,
+		Target:   target,
+		Findings: []modules.Finding{},
+	}
+
+	cdn := frameworks.DetectCDN(string(body), resp.Header)
+	if cdn == nil {
+		return result, nil
+	}
+
+	result.Findings = append(result.Findings, modules.Finding{
+		URL:      target,
+		Severity: "info",
+		Evidence: fmt.Sprintf("Fronted by %s (confidence: %.2f)", cdn.Name, cdn.Confidence),
+		Extracted: map[string]string{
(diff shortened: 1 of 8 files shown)


vmfunc (changes requested):
separate cdn pool so a cf-ray can't argmax over the real framework is the right call, and keeping every sig HeaderOnly does kill the body-badge fp. but the per-header scoping you're leaning on isn't real.

detectors/cdn.go: the `Header: "Server"`/`Header: "Via"` qualifiers do nothing, nothing reads a Signature.Header field. MatchSignatures runs through containsHeader (detector.go:126) which substring-matches the pattern against every header name and value, there's no target-header concept. so "cloudflare"/"cloudfront"/"vercel"/"netlify"/"fastly" fire on any header, not just the one the edge stamps.

concrete fp: an nginx origin with no cf-ray/x-amz-cf-id but a CSP or Link header referencing a cdn-hosted asset (cdnjs.cloudflare.com, a *.cloudfront.net img-src, vercel.app...). the brand word alone scores 0.4-0.5 -> sigmoid 0.73-0.88, past the 0.5 threshold -> you report "Amazon CloudFront" on a site that just hot-links a cloudfront bucket. referencing a cdn asset in a header is about as common as configs get.

akamai is the model here: only vendor header-name tokens (akamai-grn, x-akamai), no bare brand word, so it's clean. drop the brand-word-in-value sigs from the other five and lean on the vendor headers (cf-ray, x-fastly-request-id, x-amz-cf-id, x-vercel-id, x-nf-request-id), each clears threshold on its own at weight 1.0. or actually wire the Header field into containsHeader. right now the comment about sigs being "scoped to the specific edge-injected header" is writing a check the engine doesn't cash.

minor: body says one finding per provider but DetectCDN is single-argmax, so a cloudflare-in-front-of-vercel stack only reports one. not blocking, the fp is.

TBX3D:
the FP you sketched is exactly the right thing to be worried about here, and a
brand word matching any header would absolutely have caused it. I went to drop
the brand-word sigs and then found the engine already scopes them, so I want to
put the evidence in front of you rather than quietly not do it.

Signature.Header does get read. MatchSignatures branches on it before falling
through to containsHeader:

  if sig.HeaderOnly {
      var matched bool
      if sig.Header != "" {
          matched = headerValueContains(headers, sig.Header, sig.Pattern)
      } else {
          matched = containsHeader(headers, sig.Pattern)
      }

that is internal/scan/frameworks/detector.go:112-118 on main, and it landed in
"fix(frameworks): stop false positives and version mis-extraction (#247)" on
2026-07-02, so it was already in effect for this branch's base. containsHeader
on its own does substring-match every header name and value exactly as you
describe, which is what makes the scoping look decorative if you land on that
function first. easy read to have.

I built your concrete case as a test: an nginx origin with no cf-ray or
x-amz-cf-id, a CSP naming cdnjs.cloudflare.com, an img-src cloudfront bucket, a
vercel.app frame-ancestor and a netlify api, plus a Link preload to a cloudfront
asset. it reports nothing. a companion table keeps all six providers firing on
headers a real edge stamps. nothing was covering that before, so the scoping did
read as a comment writing a check the engine did not cash.

on the minor: agreed that DetectCDN being single-argmax means a
cloudflare-in-front-of-vercel stack only reports one. left as is since you flagged
it non-blocking, happy to take it in a follow-up.

TBX3D:
following up on my last comment, because I owe you a correction. you were right
that this had an unscoped-header false positive; I checked the wrong signatures
and told you it did not.

the brand words (cloudflare, fastly, vercel...) are scoped with Header, which is
what I verified. the vendor markers next to them were not:

  {Pattern: "cf-ray", Weight: 0.6, HeaderOnly: true}

no Header, so those did go through containsHeader and matched header values as
well as names, exactly the mechanism you described. and the value that trips it
is not exotic:

  Access-Control-Expose-Headers: CF-Ray, X-Amz-Cf-Id, X-Vercel-Id
  -> "Cloudflare" at 0.9526 on a plain nginx origin

cdns themselves emit that header, so it is close to the worst case for this
detector.

fixed by adding Signature.Presence: with Header set, the signature matches on
the header existing rather than on its value. that is the right primitive for an
edge marker, whose value is an opaque request id with nothing to match, and it
keeps the marker off header values entirely. this is the "wire the Header field
into containsHeader" option you offered, just as an explicit third mode rather
than folding it into the substring path.

covered by TestCDNIgnoresVendorHeaderNamesInValues, which fails before the
change, alongside the brand-word test from the previous round and the
true-positive table.

TBX3D:
same way:

TBX3D:
rebased on current main and wired the Header field into the signatures, which resolves the fp you flagged.

main's MatchSignatures now routes a HeaderOnly signature with `Header` set through `headerValueContains` (detector.go:114) rather than the all-headers containsHeader, so the per-header scoping is real now. every brand-word sig is scoped to its edge header (`cloudflare`->Server, `fastly`/`cloudfront`->Via, `vercel`/`netlify`->Server), and each vendor also carries a header-name presence sig (cf-ray, x-fastly-request-id, x-amz-cf-id, x-vercel-id, x-nf-request-id) that clears threshold on its own. your concrete fp is pinned by a test: an nginx origin whose CSP/Link headers reference cdnjs.cloudflare.com, a cloudfront bucket, a vercel.app frame-ancestor and a netlify api, with no edge header anywhere, stays unreported.

also fixed the body wording: DetectCDN is single-argmax, so it's one finding for the highest-confidence provider, not one per provider.

TBX3D:
reworked to make the per-header scoping real instead of decorative — you were right that the old `Header` qualifier did nothing at review time.

- wired the `Header` field into the engine (it now scopes a HeaderOnly match to that header's value via `headerValueContains`, landed on main since your review) and added a `Presence` mode for vendor markers: `matched = len(headers.Values(sig.Header)) > 0`. an edge id like `CF-Ray` carries an opaque value, so what identifies the provider is that the header was stamped at all — presence, not substring.
- brand words (`cloudflare`/`cloudfront`/`vercel`/`netlify`/`fastly`) are now scoped to the header that actually carries them (`Server`, `Via`). your exact fp — an nginx origin with a CSP/`Link` referencing `cdnjs.cloudflare.com` / a `*.cloudfront.net` asset / a `vercel.app` preview — no longer fires. pinned in `TestCDNIgnoresBrandWordsInUnrelatedHeaders`.
- vendor markers on presence also stay off values: an origin advertising `Access-Control-Expose-Headers: CF-Ray` is naming a header, not sitting behind cloudflare. pinned in `TestCDNIgnoresVendorHeaderNamesInValues`.
- addressed the copilot body-read note too: every cdn sig is HeaderOnly, so the module now drains+closes via `httpx.DrainClose` and detects on headers alone, no per-target body allocation.

on the single-argmax point (cloudflare-in-front-of-vercel reports one) — left as-is since you flagged it non-blocking; happy to split DetectCDN into a multi-result form in a follow-up if you want the stack reported.

gofmt/build/vet/golangci-lint/test all green.

PR #288, opened 2026-07-03
feat(modules): add s3, gcs and azure bucket listing modules

TBX3D:


vmfunc (approved):
the matcher shape is right. status 200 AND the structural regex AND the negative word gate on the error bodies is exactly how you avoid the 200+AccessDenied trap, and the tests cover that case directly (locked bucket returning 200 AccessDenied, 403, NoSuchBucket). anchoring the s3 regex to the s3.amazonaws.com/doc/2006-03-01 xmlns is the right call too, spaces/minio/etc all emit that same namespace so you get s3-compatible coverage for free.

one non-blocking nit: the negative word lists match anywhere in the body, so an object key named literally AccessDenied.txt (or NoSuchBucket) would suppress a real hit. edge case and the xmlns regex already does the heavy lifting, so leave it.

in.

PR #289, opened 2026-07-03
feat(recon): add localai and invokeai exposure modules

TBX3D:
detect anonymously reachable localai /system (backend and loaded-model
disclosure) and invokeai /api/v1/app/runtime_config (host path and
config disclosure). both and json key markers with status 200 and
include secured-instance negatives so 401/403 boxes are not flagged.

### DIFF
--- a/internal/modules/ai_inference_exposure_test.go
+++ b/internal/modules/ai_inference_exposure_test.go
@@ -1,0 +1,130 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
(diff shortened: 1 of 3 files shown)


vmfunc (approved):
checked both endpoints against upstream. localai /system does return backends + loaded_models[].id and ships keyless by default, invokeai /v1/app/runtime_config GET returns set_fields + config with those dir keys, unauthed by default with creds masked. markers are accurate, the quoted json keys keep the and specific, and the 401/403 negatives are covered by the status matcher. regexes are anchored, no redos. in.

PR #290, opened 2026-07-03
feat(recon): add kubeflow pipelines and metaflow exposure modules

TBX3D:
detect an anonymously reachable kubeflow pipelines apiserver, whose
same api accepts pipeline run submission with an attacker-supplied
workflow manifest (arbitrary container execution), and a metaflow
metadata service leaking flow and owner enumeration. match distinctive
snake_case json keys with status 200, fail closed on empty instances
to avoid the bare-substring false-positive class.

### DIFF
--- a/internal/modules/ml_pipeline_exposure_test.go
+++ b/internal/modules/ml_pipeline_exposure_test.go
@@ -1,0 +1,121 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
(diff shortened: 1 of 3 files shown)


vmfunc (approved):
key-triple AND'd with 200 is the right anchor on both, and the fail-closed metaflow case (empty [] carries none of flow_id/user_name/ts_epoch) is the correct tradeoff. extractors are clean, no backtracking.

one non-blocking note: the kubeflow path is v1beta1 only, so a KFP 2.x deployment on /apis/v2beta1/pipelines slips through. same response shape, so it's a one-line paths: add whenever you feel like it. not gating on it.

in.

PR #291, opened 2026-07-03
feat(info): add roundcube and zimbra webmail fingerprints

TBX3D:
fingerprint roundcube via its login-page body markers and zimbra via
its web client markers, both anded with status 200 and prose-trap
negatives to avoid matching pages that merely mention the products.


vmfunc (approved):
both matchers are anchored right. roundcube AND's the product string with the `new rcube_webmail()` bootstrap, zimbra AND's the license string with CLIENT_VERSION, so neither trips on prose that just mentions the product. nice touch on the zimbra extractor: `([^"@]...)` drops the unbuilt `@buildVersion@` placeholder instead of capturing garbage. regexes are all RE2, no backtracking to worry about. tests hit the cases that matter (prose trap, plain 200, 404, version present/absent).

in.

PR #292, opened 2026-07-03
feat(recon): add postfixadmin setup and panel fingerprints

TBX3D:
detect a reachable postfixadmin setup.php (unhardened-install signal
and info disclosure, not an auth bypass since superadmin creation is
gated on a locally configured setup password) and the login panel.
markers and the setup-password form field and login field pair with
the brand string so prose mentions do not match.

### DIFF
--- a/internal/modules/mailadmin_fingerprint_test.go
+++ b/internal/modules/mailadmin_fingerprint_test.go
@@ -1,0 +1,175 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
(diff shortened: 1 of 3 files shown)


vmfunc (approved):
field-name anchoring is the right call here. setup keys on name="setup_password" + brand, panel on the fUsername/fPassword pair, so the prose/link cases can't trip it and the tests prove it. severity split is honest too, medium info-disclosure on setup vs info on the panel, matching the setup_password gate you describe. in.

PR #293, opened 2026-07-03
feat(modules): add node inspector, rails routes and openapi exposure modules

TBX3D:


vmfunc (approved):
the double-signal gating is the right call, status + regex on the ws:// debugger url + the devtoolsFrontendUrl word means a prose mention or a bare 200 can't trip it, and the fp tests prove it. same shape on rails (Controller#Action + HTTP Verb) and openapi (spec-version regex + paths). tight.

one tiny thing, not blocking: api_version on the openapi module extracts the spec version, not info.version, so it'll read 3.0.1 instead of the api's own version. fine if that's intended, the name just reads slightly off.

in.

PR #294, opened 2026-07-03
feat(modules): add grafana tempo and otel zpages exposure modules

TBX3D:


vmfunc (approved):
both matchers are tight enough to not trip on a generic 200. tempo keys on hex traceID + rootServiceName + rootTraceName, tracez on the <title>Trace Spans</title> plus the column headers, all AND'd. no redos in any of the regexes, extractor group handling is fine, and the negative tests cover the lone-header / bare-title / 401 cases. in.

PR #295, opened 2026-07-03
feat(modules): add vaultwarden, authentik and synapse version modules

TBX3D:


vmfunc (approved):
detection is tight. all three AND the version signal against a name-specific matcher plus status 200, so the shared-endpoint twins (bitwarden on /api/config, dendrite/conduit on /_matrix/federation/v1/version) don't collide with the real thing. and the tests actually prove that, not just the happy path. the reordered-json-keys case on synapse is the right paranoia. in.

PR #296, opened 2026-07-03
feat(modules): add hpe ilo, dell idrac and redfish exposure modules

TBX3D:


vmfunc (approved):
matchers are tight the way i want them. redfish gates on the `#ServiceRoot.` odata.type anchor so a manager collection with the same RedfishVersion/Managers keys doesn't trip it, and the ilo one requires the RIMP/HSI/MP tags plus the literal 'Integrated Lights-Out' so a prose mention or a generic RIMP-shaped card can't false-positive. the negative-case tests cover exactly those. extractor regexes are all RE2 so no backtracking risk, and the `<SBSN>\s*([^<]*?)\s*</SBSN>` trim is correct. in.

PR #297, opened 2026-07-03
feat(modules): add nsq and pulsar exposure modules

TBX3D:


vmfunc (approved):
both rules lean on the right anchors. topics+start_time+version and'd is specific enough that no generic json trips the nsq rule, and pulsar_topics_count + pulsar_subscriptions_count + `^# TYPE pulsar_` won't collide with node_exporter. negative tests for 401/404/partial-body are the ones that matter and you've got them.

one thing on nsq: requiring `"health":"OK"` in the and-set means a degraded-but-exposed nsqd (health NOK) slips through, and the other three anchors already pin it to nsqd. not blocking, an exposure is an exposure even when it's unhealthy, so you might drop that matcher later. in as-is.

PR #298, opened 2026-07-03
feat(modules): add guacamole and webmin login exposure modules

TBX3D:


vmfunc (approved):
webmin's three-way and (200 + MiniServ header + "Login to Webmin" title) is the right shape, the title anchor is what keeps usermin out since it shares the miniserv header. same for guac's and-condition on guacamole-common-js + <guac-login>, prose won't trip it. usermin and lone-guac-login negative tests are the ones i'd have asked for and they're already there. in.

PR #299, opened 2026-07-03
feat(modules): add terraform tfvars, ansible vault and ci config exposure modules

TBX3D:


vmfunc (approved):
the and-chain of structural regex + vocab word + negative-html is the right shape, lines up with the maven/gradle modules. and the fp tests cover the cases that actually matter (docs page, lone assignment, plain 200, 404), which is what i care about on these.

two minor things, neither blocking:
- terraform regex needs two adjacent quoted assignments. a tfvars with a single `db_password = "..."` line, or string vars interspersed with numeric/bool ones, slips through. fine as a precision call, just know it's a real fn on the sparse case.
- ansible extractor captures only the minor digit, so vault_format_version comes out `1` for a 1.1 file. either grab `1\.[12]` whole or rename the field.

in.

PR #300, opened 2026-07-03
feat(modules): add chartmuseum and verdaccio registry exposure modules

TBX3D:


vmfunc (approved):
clean pr, detection logic holds up. chartmuseum: serverInfo is the right discriminator, helm's IndexFile emits it omitempty so a static helm repo index file drops it and only a live chartmuseum populates it. and-condition on apiVersion/entries/serverInfo means a generic yaml won't trip, and the negative html guard catches the 200 error-page case. verdaccio: content-type json plus the leading-array regex plus non-html is a solid trio, and the path is specific enough that a bare json array elsewhere won't collide. no redos, extractors are linear, nothing touches the executor. in. one note, not blocking: you fire on an empty [] too, defensible for recon since an open endpoint is the finding, just a touch noisy. fine to leave as is.

PR #301, opened 2026-07-03
feat(modules): add haproxy stats, nexus and pushgateway exposure modules

TBX3D:


vmfunc (approved):
all three are tight where it counts. the triple anchor (status + word-and + the url/success regex) means none of these trip on a bare 200, and the negatives earn their keep: the pushgateway prometheus-status test that discriminates on start_time/build_information vs startTime/CWD is exactly the right false-positive to guard against. haproxy default `/haproxy?stats` uri and the "Statistics Report for" prefix hold across versions, good call not pinning the pid-vs-hostname title. version regexes are all linear, nothing to worry about on redos. in.

PR #302, opened 2026-07-03
feat(modules): add nextcloud, jellyfin, forgejo and immich version modules

TBX3D:


vmfunc (changes requested):
the immich module is broken. extractors use `type: json` but the yaml executor only implements regex (runExtractors, executor.go:361), and the Extractor struct has no json field to unmarshal into (module.go:100). so immich_major/minor/patch never populate, the finding fires with an empty extracted map, and your own immich test (want 3 / want 1) fails against the real executor. i ran it to confirm. swap them to regex like the other three, `"major"\s*:\s*([0-9]+)` group 1 etc. verified that passes.

the other three are solid. the forgejo `+gitea-` discriminator is exactly the right call, keeps vanilla gitea (dev builds and v1.x releases) out, and the productname gating on nextcloud/jellyfin cleanly excludes owncloud/emby. checked all three against the real bodies in your tests, they hold.

fix immich and this is in.

TBX3D:
thanks for actually running it, that is the kind of review that catches real
dead modules. I went to make the swap and could not get it to fail on my end,
so let me show my working before I change anything, in case I am the one
holding it wrong.

I think the tree this was read against predates #191. the json extractor landed
on main on 2026-06-22 in "feat(modules): add json extractor (#191)". on current
main runExtractors has:

  case "json":
      part := getPart(e.Part, resp, body)
      for _, path := range e.JSON {
          if r := gjson.Get(part, path); r.Exists() { ... }

at internal/modules/executor.go:526, and Extractor carries
`JSON []string \`yaml:"json,omitempty"\`` at module.go:109. the line numbers in
the review (executor.go:361, module.go:100) match the pre-#191 tree.

rebased the branch onto current main and the immich test passes, including the
two assertions quoted in the review:

  --- PASS: TestImmichVersionExposureModule
      --- PASS: .../an_exposed_immich_version_endpoint_is_flagged_and_versioned

immich_major=3 and immich_patch=1 both populate. so I have left the extractors
as type: json rather than rewriting them to regex, since that would be a
downgrade for no gain, but say the word if you would still rather they were
regex for consistency with the other three and I will switch them.

TBX3D:
the module defines immich_major, immich_minor and immich_patch but the test only
pinned major and patch, so a regression in the middle extractor would have gone
unnoticed. added the minor assertion.

TBX3D:
same way:

TBX3D:
main resolved the immich blocker: json extractors landed in the executor (executor.go case "json", gjson-backed) and the Extractor struct now unmarshals a json field. rebased on current main, the immich test passes against the real executor, with immich_major/minor/patch populating on the flagged case. left the extractors as `type: json` since it's supported now, but happy to swap them to regex for consistency with the other three if you'd rather keep the module set uniform.

PR #303, opened 2026-07-03
fix(scan): baseline-diff lfi evidence to stop static-content false positives

TBX3D:
the lfi scan matched evidence regexes against each response with no
baseline, so a page whose static content matched a pattern (a blog
containing an /etc/passwd-shaped line, a code sample with <?php) was
reported as an lfi finding for every parameter and payload combination.
take one baseline get of the unmodified target and exclude, per
evidence class, any pattern already present in it, so only
payload-caused disclosures are counted.

### DIFF
--- a/internal/scan/lfi.go
+++ b/internal/scan/lfi.go
@@ -170,6 +170,25 @@ func LFI(targetURL string, timeout time.Duration, threads int, logdir string) (*
 
 	log.Info("Testing %d parameters with %d payloads", len(paramsToTest), len(lfiPayloads))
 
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
keying suppression on evidence.description instead of a global bool is the right call here, the poison test proves a static `<?php` baseline won't mask a real passwd leak under traversal. baseline map is written before the workers spawn and only read inside them so no race, body's closed, limitreader matches the worker read. baseline failing open (no suppression) is the correct direction for a scanner. in.

PR #304, opened 2026-07-03
fix(scan): baseline-diff sql error signatures to stop shared-template false positives

TBX3D:
checkDatabaseErrors matched db error regexes against each response with
no baseline, so a site serving a shared error template, custom 500 or
waf block page containing a db error string reported an identical
high-severity sql injection finding on every probed path. capture the
main-url response as a baseline and skip, per pattern, any error string
already present in it while still checking other patterns, so only
differential probe-caused errors are reported.

### DIFF
--- a/internal/scan/sql.go
+++ b/internal/scan/sql.go
@@ -230,8 +230,10 @@ func SQL(targetURL string, timeout time.Duration, threads int, logdir string, ca
 	}
 	wg.Wait()
 
-	// check main URL for database errors
-	checkDatabaseErrors(client, targetURL, sanitizedURL, result, logdir, &mu, seenErrors)
+	// check main URL for database errors; its body doubles as the baseline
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
per-pattern baseline diff is the right call here, not a global "any baseline hit suppresses the url". the PerPatternDifferential test is the one that matters, it pins the case where a shared mysql string and a real differential postgres leak land on the same probed path and the postgres one still fires. baseline's bounded by the 100k LimitReader, body closed, returns "" on every error path so a failed homepage fetch just degrades to no-diff. good.

one thing to be aware of, not blocking: if the homepage itself leaks say a mysql error for an unrelated reason, a genuine mysql injection gets suppressed by the baseline. it's narrow since it's per-pattern and other db types still report, and you documented it in the comment, so fine to keep as the tradeoff. in.

PR #305, opened 2026-07-03
fix(scan): gate cors http-downgrade probe on https targets only

TBX3D:
the http scheme downgrade origin was queued against every target, so on
a plain-http target the http://host probe is the target's own origin and
any correct same-origin reflection was reported as a high-severity
downgrade. mark that origin https-only and skip it when the target
scheme is not https, where a downgrade is meaningless, while still
firing it on https targets that genuinely trust the http origin.


vmfunc (approved):
right call keying the gate on the target scheme instead of the origin scheme. http://{host} against a plain-http target is just the site's own origin, reflecting it is same-origin, not a downgrade. the DowngradeFiresOnHTTPSTarget test pinning it still fires high+creds on an https target is what keeps this from turning into a false negative later. in.

PR #306, opened 2026-07-03
fix(scan): strip url whitespace before open-redirect host check

TBX3D:
pointsAtSentinel parsed the reflected location without the whitespace
and control-byte removal a browser applies, so payloads like a
space-prefixed scheme-relative host or a tab embedded in the scheme
parsed to an empty host and a real off-site redirect was missed. strip
leading and trailing c0-or-space and all tab, cr and lf before parsing,
matching the navigation-time normalization, without removing mid-string
spaces so a same-site target is never reclassified as off-site.

### DIFF
--- a/internal/scan/redirect.go
+++ b/internal/scan/redirect.go
@@ -274,6 +274,13 @@ func pointsAtSentinel(location string) bool {
 		return false
 	}
 
+	// a location like " //host" or "htt\tps://host" still navigates off-site
+	// in a browser even though it fails net/url's stricter parse.
+	location = stripURLWhitespace(location)
+	if location == "" {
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
the two-pass matches the whatwg parser, trim c0-or-space at the ends then drop tab/cr/lf everywhere, and it's before the backslash normalize + parse which is the right order. the leading-space-same-site and mid-string-space cases (redirect_test.go:48,50) are exactly the false-positives i'd have worried about, good that you pinned them. only nit: stripURLWhitespace news up a strings.Replacer every call, hoist it to a package var. not blocking.

PR #307, opened 2026-07-03
feat(modules): add dns module executor

TBX3D:
DNS modules were a stub that returned the unsupported sentinel. Implement
the executor: resolve the configured name and record type, then run the
module's matchers and extractors against the answer. A matcher targets a
part, either the record set (answer), the response status (rcode), or the
full response (default), so a status like NXDOMAIN is matchable directly.
A name's {{FQDN}} resolves to the target host. The record type and matcher
types are validated at parse time so a bad module fails to load rather than
silently matching nothing. Promotes miekg/dns to a direct dependency for the
record type codes.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -65,7 +65,7 @@ info:
 
 ### type (required)
 
-module type. `http` and `tcp` are supported.
+module type. `http`, `tcp` and `dns` are supported.
 
 ```yaml
 type: http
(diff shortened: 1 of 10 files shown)


vmfunc (approved):
solid, and the two calls that would've been easy to get wrong are both right. RawResp.String() over data.Raw in newDNSResponse is correct: retryabledns concats Raw across the retry loop (`dnsdata.Raw += resp.String()`), so keying evidence off Raw would dupe on every nxdomain retry. and answer -> AllRecords is accurate, ParseFromMsg only parses msg.Answer, so `part: answer` sees the answer section and not authority/additional. rejecting status matchers on dns at parse time is the right gate too.

one nit, not blocking: dns-spf-record spf_policy regex `v=spf1[^"]*` grabs the first quoted TXT chunk only. a long spf record split across multiple strings renders as separate quoted chunks in AllRecords, so the extractor truncates at the first boundary. the v=spf1 matcher still fires fine, it's only the extracted policy string that clips. fine for info sev, just know it.

conflicting with main, rebase it and this is in.

TBX3D:
rebased onto current main. two things needed resolving rather than replaying:

executor.go carried a stub for whichever executor was not written yet. this branch removed the dns stub, #308 removed the tcp one, so the merge kept both and neither is wanted. both are gone now, along with the shared `TestExecute*ModuleUnsupported` test.

`checkWords` grew a case-insensitive argument on main, so `checkDNSMatcher` passes `m.CaseInsensitive` through, and the docs `type` line lists http, tcp and dns.

PR #308, opened 2026-07-03
feat(modules): add tcp module executor and recon modules

TBX3D:
adds a tcp module type and executor so yaml modules can probe raw tcp
services, plus recon modules for redis, memcached, ssh, ftp, mysql, vnc,
smtp, pop3, imap and rsync. matcher semantics mirror the http executor
(matchers-condition and-default, negative inversion, word/regex/size);
status and favicon matchers are rejected at load. dial, read and write
deadlines are set and the connection is closed on every path.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -185,9 +185,10 @@ tcp:
 
 #### data
 
-an optional payload sent after connecting. use a double-quoted yaml string so
-escapes like `\r\n` are decoded before they go on the wire; a server that only
-banners (ssh, smtp) needs no data at all.
+an optional payload sent after connecting. sif decodes C-style escapes in the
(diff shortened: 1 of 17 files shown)


vmfunc (approved):
tcp executor's clean. named constants, every error wrapped and handled, decodeTCPData keeps the malformed \xHH literal instead of dropping bytes, and the ctx-deadline goroutine to unblock a pending read is the right call.

one thing i clocked, not a blocker: readTCP sets the deadline once to now+timeout and the loop only exits on a read error, so on a hold-open banner the second Read blocks the full 10s. it's bounded and only open ports pay it (closed ports never reach readTCP), so correctness is fine. drop to a short read deadline after the first data returns whenever you're next in there. shipping this.

PR #309, opened 2026-07-03
fix(scan): tolerate ref path items when parsing openapi specs

TBX3D:
paths was typed as a map of operation structs, so a spec with a $ref
string path item (valid in openapi 3.1) failed to decode and the whole
spec was dropped with no finding. decode path items into a generic tree
and extract per-operation security manually, treating a non-array
security value as absent so it inherits the global default rather than
being reported as an anonymous endpoint.

### DIFF
--- a/internal/scan/openapi.go
+++ b/internal/scan/openapi.go
@@ -79,12 +79,21 @@ const (
 // openapiSpec is the minimal slice of an openapi/swagger document we care about:
 // the version banner, info block, top-level security and the path map. unknown
 // fields are ignored by both json and yaml decoders.
+//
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
the important bit here is operationSecurity treating a non-list security value as present=false so it inherits global instead of decoding to an empty block and firing a spurious high. that's the exact edge the old nil-pointer handled and you kept it, and TestOpenAPI_NonListOpSecurityInheritsGlobal pins it. $ref fix is right too, moving the inner value to interface{} stops the string ref value from blowing up the whole document decode while /users still enumerates. in.

PR #310, opened 2026-07-03
fix(scan): grade hsts on the final response scheme after redirects

TBX3D:
hsts grading was gated on the scheme of the originally requested url,
but the client follows redirects, so an http target that redirects to
https skipped the hsts check entirely and dropped a high-severity
finding. decide the scheme from the final response request url instead,
falling back to the requested url only when no response request is set.


vmfunc (approved):
right, resp.Request.URL is the final hop and the client follows redirects by default, so an http target that 301s to https was silently dropping the hsts finding before this. grading against the final scheme lines up with reading resp.Header off that same final response too, so it's consistent. fallback and the four test cases are good. in.

PR #311, opened 2026-07-03
fix(scan): accept string-encoded jwt exp claim

TBX3D:
numericClaim only accepted exp as a json number, but some emitters
serialize it as a quoted string. a token with a past string exp was
reported as missing exp instead of expired, an expiry false negative
and a missing-claim false positive on the same token. also parse a
string value via strconv.ParseFloat, falling back to absent when it
does not parse.

### DIFF
--- a/internal/scan/jwt.go
+++ b/internal/scan/jwt.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"net/http"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
numericClaim only feeds the exp check at jwt.go:282, so widening it to string-encoded NumericDate is exactly the right scope, and the fallback-to-absent on unparseable keeps the missing-exp path honest. test at jwt_test.go:78 pins both halves (flags expired, doesn't double-report missing). in.

PR #312, opened 2026-07-03
fix(report): map sarif level to finding severity

TBX3D:
sarif results were hardcoded to level "warning", so a critical finding
and an informational one landed identically in code-scanning ingests.
carry the normalized severity onto report.Result and map it: critical
and high to error, medium to warning, low and info to note. the severity
is derived by flattening each module result and taking the worst rank,
and an absent severity keeps the old "warning" default.

### DIFF
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -18,9 +18,13 @@ package report
 import "encoding/json"
 
 // Result is one module's output for one target. Data is whatever the scanner
-// returned, carried as raw json so report stays free of scan types.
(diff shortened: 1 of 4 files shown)


vmfunc (approved):
clean. reportSeverity taking the worst rank across the flattened findings is the right call for a per-module sarif result, and the ordering in finding.Severity (unknown < info < ... < critical) makes the plain `>` comparison correct. empty -> warning keeps the old default so nothing regresses. the explicit `case "medium"` is redundant with the default but reads fine as intent. in.

PR #313, opened 2026-07-03
feat(report): add json findings report with -json

TBX3D:
adds a `-json <path>` flag that writes the run's normalized findings as a json array, one object per finding with its target, module, severity, key, title and evidence. sarif and markdown were the only structured exports and both drop the per-item detail the finding model already carries.

detection confidence is threaded onto the finding model so framework detections report their score; it rides into the json report and is omitted for scanners that carry none.

the `-silent` line format and the diff snapshot on-disk shape stay unchanged: the report uses a dedicated json view rather than marshalling the finding struct directly.

verified end to end by running a scan against a local server with `-json out.json` and parsing the result, plus unit tests for the json shape, omitted-zero-confidence, and the confidence threading.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -69,6 +69,7 @@ type Settings struct {
 	Probe             bool
 	SARIF             string // path to write a sarif 2.1.0 report to ("" = off)
 	Markdown          string // path to write a markdown report to ("" = off)
+	JSONReport        string // path to write a json findings report to ("" = off)
 	Silent            bool   // route chrome to stderr, print one finding per line to stdout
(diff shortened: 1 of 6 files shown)


vmfunc (approved):
dedicated reportFinding view instead of marshalling Finding directly is the right call, keeps the snapshot on-disk shape and -silent line format from moving. writeJSONReport mirrors writeReports cleanly, and JSONReport never returning nil (make with 0 cap -> "[]") is the correct guard. confidence threading off r.Confidence with omitempty for the recon findings that carry none is right. in.

PR #314, opened 2026-07-03
feat(frameworks): report every detected technology

TBX3D:
framework detection collapsed to a single argmax winner, so a page built from several technologies (react behind a next.js server, say) reported only one and dropped the rest.

adds `DetectFrameworks`, which keeps every detector hit above the threshold ranked most-confident first, and wires the `-framework` path to emit one finding per detected framework. the single-winner `DetectFramework` stays for callers that want just the top match; both share one fetch through a new `gatherDetections` helper, so the extra coverage costs no additional request.

verified live against a local server carrying both next.js and express markers: `-framework` now reports both `Next.js` and `Express.js` where it previously reported one. new tests cover the multi-report ordering and the no-match case; the existing single-detection tests are unchanged.


vmfunc (approved):
clean split. gatherDetections sharing one fetch between the single- and multi-winner paths is the right call, and moving spin.Stop() to a defer there quietly kills the missing-stop paths the old code had on every error return.

threshold semantics stay consistent (> detectionThreshold in both), the sort tie-break on name keeps output stable, and since the registry is name-keyed there's no dup-framework risk in the multi path. no new false-positive surface, each hit still gates on the same threshold.

conflicts with main though, rebase and i'll merge.

PR #316, opened 2026-07-03
feat(modules): embed built-in modules in the binary

TBX3D:
modules loaded only from an exe-adjacent `modules/` dir or `~/.config/sif/modules`, so a bare `go install`ed binary shipped zero modules and silently found nothing to run.

embeds the `modules/` tree and falls back to it when no filesystem builtin dir is found. the on-disk builtin dir and the user-override dir still take precedence, so a dev tree and a release that ships the folder alongside the binary keep working unchanged.

the embed directive lives at the repo root because `go:embed` cannot reference a parent directory (`modules/` sits above `internal/modules` where the loader is); it registers the embedded fs with the loader through a small package hook to avoid an import cycle back into the root package.

verified: `--list-modules` from a directory with no `modules/` present lists all 192 built-ins (from the embed), and from the repo root it lists the same 192 off the filesystem with no double-loading.

### DIFF
--- a/embed_modules.go
+++ b/embed_modules.go
@@ -1,0 +1,40 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
+:   █▀ █ █▀▀   ·   Blazing-fast pentesting suite                                :
+:   ▄█ █ █▀    ·   BSD 3-Clause License                                         :
+:                                                                               :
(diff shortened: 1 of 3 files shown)


vmfunc (approved):
clean. the `before := l.loaded` capture is the right way to gate the embed fallback so on-disk and embedded never double-register, and routing the fs through SetBuiltinFS keeps internal/modules from importing the root package back. init() runs because cmd/sif/main.go pulls in the root pkg, so the registration actually fires.

one tiny thing, not blocking: `//go:embed modules` also pulls the .go script files into the binary since loadFS only reads yaml. dead weight, ignore it or add a yaml-only glob later.

in.

PR #317, opened 2026-07-03
feat(modules): add dynamic-variable request chaining

TBX3D:
http modules could only fire independent one-shot requests: an extractor recorded a value for reporting, but nothing could feed it back into a later request, so multi-step flows (fetch a token, then use it) were inexpressible.

adds a `requests:` chain where steps run in order sharing a variable map. each step's extractors populate `{{name}}` references usable in the path, headers and body of later steps. a step with matchers records a finding on a match and halts the chain on a miss, so a login or setup step can gate an authenticated follow-up.

scope is deliberately tight: an ordered request list with variable propagation, not a full workflow/DSL engine. the single-request form is untouched and stays the concurrent default when no chain is defined.

```yaml
http:
  requests:
    - path: "{{BaseURL}}/login"
      extractors:
        - type: json
          name: token
          json: ["token"]
    - path: "{{BaseURL}}/secret"
      headers:
        Authorization: "Bearer {{token}}"
      matchers:
        - type: word
          part: body
          words: ["TOP SECRET"]
```

tests drive the real executor through the shared http client against a two-endpoint httptest server: the chain reaches the protected step only because the extracted token crosses into the second request's header, and it records nothing when the token is missing or an earlier step's matcher fails. the yaml schema round-trip is pinned too. production chain modules are a natural follow-up.


vmfunc (approved):
chain logic itself is clean. bodies closed after each step, ctx.Err() checked per iteration, snapshotVars keeps each finding from aliasing the vars map that later steps keep mutating, and sequential-by-design is the right call for a value-dependency chain. the matcher-gates-and-halts semantics match the description.

thing is, this doesn't stand on main. the tests build a json extractor (Extractor.JSON, type: json) and on main runExtractors only handles regex and the Extractor struct has no JSON field. same for HTTPConfig.MatchersCondition, validateMatchers/validateMatchersCondition, and generateHTTPRequests returning an error, all referenced in the diff but not present on main yet. so this sits on top of the json-extractor + matchers-condition + request-error PRs in your stack and has to merge after them, with a rebase onto main once they're in. green CI here is against the stack tip, not main.

nit: substituteVariablesWithVars iterates vars in map order and replaces literal {{name}}, so an extracted value that itself contains {{othervar}} can get re-substituted depending on order. server-controlled and marginal, but worth a guard if you ever chain untrusted extracts forward.

in once the base PRs land.

PR #318, opened 2026-07-03
fix(modules): route module scans through the shared http transport

TBX3D:
module execution built a bare http client, so -proxy, -H, -cookie and
-rate-limit applied to every other scanner but silently not to module
scans. route through httpx.Client so modules share the configured
transport like the rest of the run.


vmfunc (approved):
yep, modules were the last path still building a bare client while everything else goes through httpx.Client. the Options.Client field and the nil fallback in executor.go were already there, this just wires it, so -proxy/-H/-cookie/-rate-limit finally hit module scans via the shared roundtripper. correct.

conflicting with main, rebase and it's in.

PR #319, opened 2026-07-03
feat(modules): add case-insensitive word matcher option

TBX3D:


vmfunc (approved):
folds both content and needle, default behavior unchanged, and the AdMiN/admin test pins the exact case i'd worry about. in.

PR #320, opened 2026-07-03
fix(frameworks): match cve entries against a bare-major version

TBX3D:
drupal's generator meta only exposes a bare major, so the extracted
version is "10"/"9". versionAffected only matched when the affected
entry was a component-prefix of the version, never the reverse, so a
bare major never matched a dotted affected entry and every drupal cve
was silently missed. match on dotted boundaries in either direction.

### DIFF
--- a/internal/scan/frameworks/cve_internal_test.go
+++ b/internal/scan/frameworks/cve_internal_test.go
@@ -65,6 +65,18 @@ func TestResolveVersionFeedsCVELookup(t *testing.T) {
 	}
 }
 
+// Drupal's generator meta only exposes a bare major, so the extracted version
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
confirmed against version.go:116, the drupal generator regex only captures `(\d+)` so the detected version really is a bare major and the old one-directional prefix dropped every drupal cve. reverse prefix is the right fix and stays on dotted boundaries, the "1" vs "10.0" and "10" vs "100.0" tests are exactly the cases i'd have worried about. blast radius is contained to the bare-major detectors and flagging coarse-on-major is the correct conservative call for a scanner. in.

PR #321, opened 2026-07-03
fix(scan): stop one over-long url from dropping the wayback feed

TBX3D:
the wayback line scanner capped each line at 1mb, so a single archived
url with a huge query string made scanner.Err return ErrTooLong and the
whole feed was discarded. read lines without a per-line cap, still bounded
by the passiveMaxBytes read limit.


vmfunc (approved):
yep, scanner.Err returning ErrTooLong nuking the whole feed on one fat line was a real bug. ReadString('\n') handles the trailing line-without-newline case right and overall size is still capped by passiveMaxBytes. in.

PR #322, opened 2026-07-03
fix(frameworks): stop prose and image mime types from faking magento

TBX3D:
the bare "Magento" body word matched any page that merely named the
platform, and "mage/" is a substring of the near-ubiquitous "image/"
mime type, so a migration guide carrying a single image reference
scored both signatures and cleared the detection threshold. key on the
frontend static-asset base, the data-mage-init widget attribute and the
module namespace, which only appear when magento actually rendered the
page.


vmfunc (approved):
diagnosis is right, "image/" contains "mage/" as a substring and bare "Magento" caught any page that named it. new anchors are all structural. checked against detector.go: totalWeight 1.4, sigmoid centers at 0.3, so a single 0.4 marker lands at 0.286 and won't trip, you need two real signals to clear the threshold. good fp resistance and recall holds since real magento carries data-mage-init + /static/frontend/ + Magento_ together. in.

PR #323, opened 2026-07-03
fix(js): bound next.js manifest fetch with the scan timeout

TBX3D:
GetPagesRouterScripts fetched _buildManifest.js with httpx.Client(0),
which sets no client timeout, so a slow or hostile manifest host could
hang the whole scan on the response read. thread the caller's scan
timeout through instead, matching every other fetch in the package.

### DIFF
--- a/internal/scan/js/frameworks/next.go
+++ b/internal/scan/js/frameworks/next.go
@@ -29,6 +29,7 @@ import (
 	"net/http"
 	"regexp"
 	"strings"
+	"time"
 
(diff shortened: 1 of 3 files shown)


vmfunc (approved):
yep, this was the last straggler in the package still on httpx.Client(0) while JavascriptScan itself and ScanSupabase already thread timeout. Client.Timeout covers the body read so this actually kills the hang, not just connect. test with the blocking handler + release chan is the right shape. in.

PR #324, opened 2026-07-03
fix(scan): honour the caller context when downloading the ports list

TBX3D:
the common-scope ports fetch built its request with context.TODO(),
so a cancelled run or -max-time deadline could not abort a hung list
download even though the dialer already used the passed context.


vmfunc (approved):
ports.go:50 was the only spot still on context.TODO() while the dialer already threaded ctx, so the list download couldn't be aborted mid-flight. right fix. and the test cancelling before the call so only ctx-awareness can unblock it is exactly the thing that proves it. in.

PR #325, opened 2026-07-04
feat(modules): add encoded word matcher option

TBX3D:
word matchers gain an optional encoding of hex or base64. when set, each
configured plaintext word is rendered into that encoding before matching,
so a marker that appears encoded in a response (a base64 blob, a hex dump)
can be matched by its known plaintext form. an unset encoding keeps the
existing literal behaviour byte for byte. an unknown encoding value is
rejected at module load.

### DIFF
--- a/internal/modules/executor.go
+++ b/internal/modules/executor.go
@@ -15,6 +15,8 @@ package modules
 import (
 	"bufio"
 	"context"
+	"encoding/base64"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


vmfunc (approved):
clean. unset path stays byte-for-byte and the unknown-encoding reject reuses the validateMatchers load-time gate, exactly where it belongs. good reuse of encodeMatcherWord("") as the validator too.

two sharp edges to note in the module docs, both false-negative only so not blocking: base64 goes through StdEncoding, so it only matches when the marker is 3-byte aligned in the source blob, and it won't catch base64url (-_ , the JWT/web variant). and hex.EncodeToString is lowercase while checkWords is case-sensitive, so an uppercase hex dump slips past. fine for the opt-in heuristic, just don't want a rule author expecting it to catch a marker mid-blob.

in.

TBX3D:
rebased. main added the range matcher and word case-insensitivity in the meantime, so the merge is a union rather than a replay: `validateMatchers` is one switch over favicon, range and word instead of the two ifs it grew from, `Matcher` carries `Encoding` alongside `Source`/`Min`/`Max`/`CaseInsensitive`, and the word matcher encodes first and then passes `m.CaseInsensitive` to `checkWords`. the merge commit is gone, so this is a single commit now.

vmfunc:
clean little feature, nicely tested. it refactors validateMatchers though, which just changed under #382, so it needs a rebase. two small things while you're in there: add a docs/modules.md line for the `encoding` field, and note the base64 caveat (a marker embedded mid-stream in a bigger base64 blob won't match due to alignment/padding). rebase + those and it's in.

---

PR #326, opened 2026-07-04
fix(scan): truncate shodan banners on rune boundaries

TBX3D:
truncateBanner sliced by byte offset, so a multibyte utf-8 rune landing
across the limit was cut mid-sequence and emitted invalid utf-8. shodan
banners routinely carry non-ascii bytes, so this corrupted logged and
printed output. count runes instead so the cut lands on a boundary.


vmfunc (approved):
correct fix, byte slicing at maxLen was splitting multibyte runes on hostile banner data. maxLen becomes a rune count instead of a byte count now, which is what you actually want for banner display. good test cases too. in.

PR #327, opened 2026-07-04
fix(scan): percent-encode dork queries before search

TBX3D:
the google-search client only swaps spaces for '+' and drops the term
into the query string verbatim, so a dork carrying a raw '#' or '&' cut
the request url short at the fragment or split it into stray query
params. the search then ran against a fragment of the intended dork.
encode the term up front so the whole dork survives.

### DIFF
--- a/internal/scan/dork.go
+++ b/internal/scan/dork.go
@@ -18,8 +18,8 @@ package scan
 import (
 	"bufio"
 	"context"
-	"fmt"
 	"net/http"
+	"net/url"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
checked the actual client, buildUrl in rocketlaunchr/google-search just does Trim then space->'+' and drops the term straight into ?q=, zero encoding. so yeah, a raw '#' landed in the fragment and '&' split the params, exactly as you describe. QueryEscape is the right call here: space still ends up as '+' so the client's own replace is a no-op, and '#'/'&'/'+' round-trip clean on google's side. nice that buildSearchURL in the test actually mirrors buildUrl instead of asserting against itself, the truncation guard is real. in.

PR #328, opened 2026-07-04
fix(frameworks): stop django source prose from faking detection

TBX3D:
the django detector keyed on the import paths django.contrib and
django.core, which appear only in tutorials and settings snippets and
never in a rendered response, so any two of them crossed the threshold
and fingerprinted a docs page as a live django server. replace the
source tokens with real response artifacts: the csrfmiddlewaretoken
hidden field, the csrftoken cookie and the /static/admin/ asset path.


vmfunc (approved):
good diagnosis. django.contrib/django.core only ever show up in source snippets, and two of them cleared 0.3 so any settings.py in a <pre> got fingerprinted as live django.

nice catch that the old csrfmiddlewaretoken sig was HeaderOnly:true, so it never actually matched (it's a body form field, never a header). dropping name="csrfmiddlewaretoken" to the body and swapping in /static/admin/ makes it a real artifact instead of dead weight. csrftoken alone still can't trip it (0.3 == 0.5 confidence, not >), so the cookie-only case stays quiet. good.

in.

PR #329, opened 2026-07-04
fix(report): sort sarif rules for reproducible output

TBX3D:
the rules list was built by iterating a set, so the same scan produced
sarif with the rules in a different order on every run, defeating diffing,
caching and snapshot tests. sort the ids before emitting them.

### DIFF
--- a/internal/report/report_test.go
+++ b/internal/report/report_test.go
@@ -13,6 +13,7 @@
 package report
 
 import (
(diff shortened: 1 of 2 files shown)


TBX3D:
closing as a duplicate of #337, which applies the identical sarif rule-id sort fix and also covers the markdown crlf fix.

PR #330, opened 2026-07-04
refactor(cli): drop comments restating adjacent dns/subdomain code

TBX3D:
two comments in scanTarget restate the line they sit above: `dnsResults = result // Store the DNS results`, and a `// Only run subdomain takeover check if DNS scan is enabled` above the `if app.settings.SubdomainTakeover` it describes. drop both.

this PR previously carried the scanTarget/targetScan extraction as well. that landed on main in #367 with the same signature and the same `sif_scantarget_test.go`, so only the comment trim is left here.

### DIFF
--- a/sif.go
+++ b/sif.go
@@ -472,11 +472,10 @@ func (app *App) scanTarget(url, storeDir string, wantReport bool) (targetScan, e
 			log.Errorf("Error while running dns scan: %s", err)
 		} else {
 			moduleResults = append(moduleResults, ModuleResult{"dnslist", result})
-			dnsResults = result // Store the DNS results
+			dnsResults = result
 			scansRun = append(scansRun, "DNS Scan")
 		}
 
-		// Only run subdomain takeover check if DNS scan is enabled
 		if app.settings.SubdomainTakeover {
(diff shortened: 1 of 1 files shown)


vmfunc (approved):
clean extraction. the only spot that isn't mechanical is the apimode marshal-failure path, and you got it right: `return targetScan{scansRun: scansRun, logFiles: logFiles}, nil` keeps the scan labels and log files but drops findings/report, same as the old `continue`. logFiles moving to a local and merging back in Run still lands in app.logFiles before PrintSummary, so the summary is unchanged, and nothing reads app.logFiles mid-scan so the timing shift is invisible.

TestScanTargetIsolatesPerTargetState is the right test for the seam. two sequential targets each reporting a single scan is exactly what a shared slice would break.

conflicts with main though, rebase and it's in.

TBX3D:
rebased, and there is almost nothing left. #367 landed `scanTarget`/`targetScan` with the same signature and the same `sif_scantarget_test.go`, so the extraction this PR proposed is upstream already. only the two-line comment trim survives; retitled and rewrote the description to match.

that means your approval no longer covers what is here, so please re-read before merging, or close it if a two-comment cleanup is not worth the churn. either is fine by me.

PR #331, opened 2026-07-04
feat(output): add routable Sink for per-scan output capture

TBX3D:
introduce a Sink (writer plus an interactive flag) that the package-level
chrome funcs delegate to via DefaultSink, and give ModuleLogger a sink it
writes through. a caller can now hand a scan its own Sink so that scan's
chrome lands on a dedicated writer instead of the process-wide default.
default routing is unchanged: Info/Warn/Module and friends still write to
the same stdout/stderr sink SetSilent configures.

### DIFF
--- a/internal/output/output.go
+++ b/internal/output/output.go
@@ -188,7 +188,7 @@ func (s *Sink) Writer() io.Writer { return s.w }
 // buffer sink reports false so spinners and progress bars stay silent.
 func (s *Sink) Interactive() bool { return s.interactive }
 
-// Info prints an informational message with [*] prefix.
+// Info logs a [*]-prefixed message; a no-op in API mode.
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
clean. the Interactive()/!silent gate lines up with ClearLine so a captured buffer sink won't try to animate, and the isolation test (TestSinkDoesNotLeakToDefault) covers the thing that actually matters for per-target capture.

one thing to keep in the back of your head: Module() now snapshots the writer at construction via DefaultSink(), where the old ModuleLogger read the global sink at write time. package funcs still track it live. today it's fine because SetSilent runs once at startup before any output.Module() call, but if we ever toggle silent mid-run, loggers built before the toggle keep the old writer. not blocking, just noting it. in.

PR #332, opened 2026-07-09
fix(httpx): stop cookie/auth header leak on cross-host redirect

TBX3D:
sif installs a custom CheckRedirect for crawl-scope enforcement, which
bypasses net/http's own redirect header stripping. roundTripper.RoundTrip
now mirrors that stripping itself: Cookie and Authorization are dropped
whenever a redirect crosses to a different host, matching what stdlib
would do without the custom hook. same-host redirects (including a
login bounce or trailing-slash normalization) keep working unchanged.

tests cover both directions: a cross-host redirect must not see the
original cookie/auth header, and a same-host redirect must still see it.

### DIFF
--- a/internal/httpx/httpx.go
+++ b/internal/httpx/httpx.go
@@ -229,6 +229,17 @@ func parseHeaders(raw []string) (map[string]string, error) {
 	return headers, nil
 }
 
+// sensitiveHeaders mirrors net/http's own redirect header policy (see
+// shouldCopyHeaderOnRedirect / copyHeaders in net/http/client.go): these are
+// the headers the stdlib strips itself when a redirect crosses hosts, so
+// injecting them back in at the transport layer would defeat that stripping.
+var sensitiveHeaders = map[string]bool{
+	"Cookie":           true,
+	"Cookie2":          true,
+	"Authorization":    true,
+	"Www-Authenticate": true,
+}
+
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
logic is right. `req.Response.Request.URL.Host != req.URL.Host` is the correct signal, and re-injecting per hop is exactly why the strip has to live in the transport: the client's header copier only sees the initial request headers, never what roundTripper injects, so it can't strip them itself. that's the real gap, good catch.

the test design is the part i want to call out. httptest servers are all 127.0.0.1 and net/http reads distinct ports on one ip as same-host, so the only way to actually cross a host boundary is to remap the names at DialContext the way hostMappingTransport does. most attempts at this test silently never cross and pass for the wrong reason. this one doesn't.

one thing before it goes in: this is cut against the httpx hardening branch, not main. the roundTripper struct alignment differs (your base has a field main doesn't) and the RoundTrip in your diff has no limiter.Wait at the top, so it won't apply on current main, git apply fails. rebase onto main and make sure the rate-limiter block stays at the top of RoundTrip after the merge, since your crossHostRedirect lines land in the same spot.

also fine to leave: it's stricter than stdlib, exact host match so foo.com -> sub.foo.com strips where net/http would keep. that's the safe direction, no objection, the comment just says it mirrors stdlib when it's a bit tighter.

rebase and confirm the limiter survives, then in.

PR #333, opened 2026-07-09
fix(crawl): stop redirects escaping crawl scope, cap fanout

TBX3D:
colly's AllowedDomains check is hostname-only, so a redirect to the same
host on a different port (a cloud metadata endpoint on 169.254.169.254,
say) was treated as in-scope. the custom CheckRedirect now checks host:port
exactly and caps redirect hops, since installing a custom CheckRedirect
drops net/http's own default hop limit.

the second commit caps MaxRequests, which colly left at 0 (unlimited);
a hostile site could otherwise make a depth-bounded crawl request an
unbounded number of pages. the third commit corrects a doc comment that
claimed robots.txt was honored; it never was, and this scanner's use case
(authorized recon) intentionally ignores it.

### DIFF
--- a/internal/scan/crawl.go
+++ b/internal/scan/crawl.go
@@ -36,6 +36,11 @@ func (r *CrawlResult) ResultType() string { return "crawl" }
 // compile-time check so a result-type drift fails the build, not a run.
 var _ ScanResult = (*CrawlResult)(nil)
 
+// maxCrawlRequests caps the total pages a single Crawl call will fetch, so a
+// hostile page fanning out to an attacker-controlled number of links can't
+// turn a bounded-depth crawl into an unbounded one (b^d).
+const maxCrawlRequests = 2000
+
 // Crawl spiders the target up to depth, following same-host links/scripts/forms.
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
the fix is right, but the comment undersells the hole. SetClient replaces colly's client wholesale (colly.go:1050 is just `c.backend.Client = client`), so colly's own CheckRedirect never ran here at all. the old httpx client had a nil CheckRedirect, which means net/http's default kicked in and followed redirects to *any* host, not just same-host-different-port. so pre-fix an in-scope page 302'ing to evil.com or 169.254.169.254 actually got fetched. your host:port check closes that. worth fixing the comment in crawl.go and crawl_test.go to say so, right now it reads like colly was doing a hostname check it wasn't.

rest is correct: the hop re-cap mirrors net/http's default so you don't silently drop it, MaxRequests bounds colly's own Visits and redirect hops don't count against it (which is what you want), and the reject path leaks nothing since net/http closes the last response itself before handing back the error. the robots test pinning the intentional not-honored behavior is a good call, that comment was flat wrong before. in.

PR #334, opened 2026-07-09
fix(notify): neutralize markdown injection and redact webhook url from errors

TBX3D:
attacker-controlled finding content (a page title, a crawled url) reached
the slack/discord code block verbatim. a title embedding a closing triple-
backtick fence could break out of the wrapping block and inject live
markdown; slack additionally resolves a bare "<...|...>" as a link/mention
independent of code-fence boundaries, so the three slack control
characters are now entity-escaped ahead of fencing.

the second commit redacts the webhook url from transport errors: for
slack/discord/generic sinks the url itself is the credential, and
http.Client wraps every transport failure in a *url.Error that quotes it
verbatim. errors are now unwrapped to the underlying cause and re-prefixed
with just the host, so a failure is still debuggable without leaking the
secret into a log line.

sanitizeFence breaks every backtick rather than exact triples. replacing only
triples leaves a trailing bare backtick, so a run of 5, 8 or 11 backticks
reformed a contiguous triple and reopened the fence; discord takes the sanitized
body with no entity escaping, so that was a live breakout for mentions and
masked links. the breakout test walks 3, 4, 5, 6, 8 and 11.

### DIFF
--- a/internal/notify/message.go
+++ b/internal/notify/message.go
@@ -76,15 +76,11 @@ func postJSON(ctx context.Context, client *http.Client, endpoint string, payload
 	return nil
 }
 
-// redactTransportErr strips the request url out of a client.Do failure before
-// it reaches a log line or a returned error. for these providers the url IS
-// the credential (a slack/discord/generic webhook secret is the whole url; a
-// telegram bot token rides the path) and http.Client wraps every transport
-// failure in a *url.Error whose Error() quotes that url verbatim. the
-// underlying cause (*net.OpError, a dns error, tls, timeout, ...) only ever
-// mentions the host:port, never the path, so unwrapping to it and letting the
-// caller prefix the host separately keeps a transport failure debuggable
-// without leaking the secret.
+// redactTransportErr strips the webhook url out of a client.Do failure. for
+// these providers the url IS the credential, and http.Client wraps every
+// transport failure in a *url.Error whose Error() quotes it verbatim; unwrap
+// to the underlying cause (which only ever mentions host:port) and let the
+// caller prefix the host separately.
 func redactTransportErr(err error) error {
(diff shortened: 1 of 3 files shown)


vmfunc (changes requested):
sanitizeFence has a bypass. the replacement `​`​` ends on a bare backtick, so any backtick run whose length is ≡2 mod 3 reforms a clean fence: a title of 5 backticks (or 8, 11...) comes out with the trailing three contiguous again. i checked it, n=5/8/11 all give 3 adjacent backticks back. discord takes sanitizeFence with no entity escaping, so that's a live breakout, @everyone and [x](url) render outside the block. slack only survives it because you escape <>& separately.

TestNotifyCodeBlockBreakoutNeutralized only exercises n=3 and asserts fences<=2, so it's green while the hole is open. drop in an n=5 case and it fails.

fix is to not leave a trailing bare backtick. simplest is to break every backtick instead of just triples: strings.ReplaceAll(body, "`", "`"+zwsp). no run of 2+ can survive that.

the rest is right: unwrapping the *url.Error and re-prefixing req.URL.Host keeps host:port and drops the path token for all four sinks, and the slack <>& escaping is correct. close the fence bypass and this is in.

TBX3D:
confirmed, and the n=5 case fails exactly as you said. the old
TestNotifyCodeBlockBreakoutNeutralized only drove n=3, so it stayed green:

  INJECTION: run of 5 backticks added 2 extra code fences, breaking out:
  "...`​`​```\n@everyone pwned <https://evil.test|click>\n`​`​```..."

took your fix, ReplaceAll on the single backtick, so no two are ever adjacent
and no run length can reform a triple. the breakout test now walks 3, 4, 5, 6,
8 and 11 rather than the one case.

PR #335, opened 2026-07-09
fix(report): dedup header findings by key with deterministic sort

TBX3D:
header findings were deduped by name alone, so two Set-Cookie (or other multi-valued) headers with different values collapsed onto one finding.

the key now disambiguates by how many times the header name has been seen, not by the value. Key is documented to be run-stable (finding.go:41), and folding the value in would destabilize every volatile single-valued header: Date rides on nearly every response, alongside ETag, Age, CF-Ray and X-Request-Id, so each scan would mint a fresh key and the diff/notify layer would report them added+removed every run.

the count is per name rather than the slice index: headers.go ranges over resp.Header, so a name's slice position is randomized per run while the order of values within one name is preserved.

Flatten's output is also sorted by key, since several flatten paths (in particular the JS env-var map) source from randomized Go map iteration and produced a different finding order on every run.

### DIFF
--- a/internal/finding/finding.go
+++ b/internal/finding/finding.go
@@ -20,6 +20,7 @@ package finding
 import (
 	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/projectdiscovery/nuclei/v3/pkg/output"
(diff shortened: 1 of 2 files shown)


vmfunc (changes requested):
the sort in Flatten plus the js env-var stability test are the right call, that map iteration needed pinning and the 200-run test is a good way to lock it.

the header key change is the wrong fix though. folding the full value into the key breaks the run-stable identity the Key field is documented to guarantee (finding.go:39, "the same finding across two runs produces the same Key"). the collapse you're actually hitting is only repeated header *names* (Set-Cookie), but headers.go emits every resp.Header, so you're now also destabilizing every single-valued volatile header: Date is on basically every response, plus ETag/Age/Server-with-version/CF-Ray/X-Request-Id. all of those were name-stable before and now churn a new key each run, so the diff/notify layer that keys off Finding.Key reports them added+removed on every scan of every target.

disambiguate by occurrence index within the same name instead, e.g. key("headers", h.Name+"#"+strconv.Itoa(i)). splits the two Set-Cookies, leaves Date and friends stable across runs. adjust the test's wantKey accordingly and it's in.

TBX3D:
right, the value was the wrong axis and Date/ETag/Age/CF-Ray would have churned
a fresh key every run. added TestVolatileHeaderKeyIsRunStable, which fails on
the old behaviour with two Date values one second apart.

one adjustment to what you sketched: the disambiguator counts occurrences within
the name rather than using the slice index. headers.go builds its slice with
`for name, values := range resp.Header`, so a header's slice position is
randomized per run and key("headers", h.Name+"#"+strconv.Itoa(i)) would churn on
any multi-header response. the order of values within one name is preserved, so
a per-name counter is stable. TestHeaderKeyIndependentOfEmissionOrder pins that
by flattening the same headers in two different orders.

the two Set-Cookies still split, Date and friends stay put, wantKey updated.

TBX3D:
one correction on my own fix: the occurrence suffix used "#", and "#" is a valid
header field-name character under rfc7230, so a real header named "Foo#1" took
the key of the second "Foo".

switched the separator to ":", which rfc7230 excludes from a field-name, so the
suffix cannot collide with any real header. TestHeaderOccurrenceSuffixCannotCollide
covers it and fails on the old separator.

TBX3D:
same way:

PR #336, opened 2026-07-09
fix(output): sanitize control bytes and truncate rune-safe before terminal print

TBX3D:
several code paths wrote to stderr/stdout directly instead of through the
output sink: module loggers, the whois record dump, and a couple of scan
status lines. under -concurrency>1, only the sink is lock-wrapped, so a
raw fmt.Println/log write could interleave or tear with everything else.
those now route through output.Info/Error.

Sanitize also strips terminal control bytes (OSC, non-SGR CSI, C0/C1,
DEL) from attacker-controlled content before it reaches the terminal,
while keeping \t/\n and SGR color codes intact so styling still works.
progress-item truncation is now rune-aware so a multibyte character
straddling the cut point isn't split into invalid UTF-8.

### DIFF
--- a/internal/output/nocolor_test.go
+++ b/internal/output/nocolor_test.go
@@ -1,0 +1,26 @@
+package output
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/charmbracelet/lipgloss"
+)
+
+// regression guard: Highlight.Render must strip embedded control/OSC bytes
+// from its argument before styling it, even under NO_COLOR, so an
+// attacker-controlled string (a response title, header, ...) can't rewrite
+// the terminal title or move the cursor when it's highlighted for display.
+func TestHighlightRenderStripsEmbeddedESC(t *testing.T) {
+	t.Setenv("NO_COLOR", "1")
+	lipgloss.SetColorProfile(4)
+	evil := "\x1b]0;PWNED\x07x"
+	out := Highlight.Render(evil)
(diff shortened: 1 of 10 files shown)


vmfunc (approved):
the split is right: sanitize at the print boundary, leave extractTitle / ProbeResult.Title / logger.Write byte-accurate so the json + file record stay true and only the terminal gets scrubbed. probe_ansi_test pins exactly that.

sanitize keeping only m-final CSI is correct, sgr color survives, cursor/clear/osc dropped, and it's a linear scan so no redos. output.Info is printf-style so `output.Info("%s", result)` not `output.Info(result)` in whois was the right call, keeps a % in a whois record from being eaten.

one nit: Style.Render is a pointer receiver. every call site today is an addressable package var so it's fine, but a non-addressable Style (map elem, or one assigned into an interface) would bind the promoted lipgloss Render and skip sanitize silently. make it a value receiver, nothing here mutates and it closes that off.

conflicting with main, rebase and it's in.

TBX3D:
rebased. output.go was restructured around the `Sink` type in the meantime, which made this smaller: the package-level `Info`/`Success`/`Warn`/`Error` all delegate to `DefaultSink()` now, so `Sanitize` goes in the four `Sink` methods and the four `ModuleLogger` methods instead of the eight separate call sites it used to patch. the `Style` wrapper over `Highlight`/`Muted`/`Status`/`Severity*` is unchanged, as are the whois, module-logger and progress-truncation commits.

PR #337, opened 2026-07-09
fix(report): sort sarif driver rules and strip crlf from markdown headers

TBX3D:
SARIF's rule set is collected from a map, so without an explicit sort the
emitted driver.rules order (and therefore the report bytes) varied
run to run for an identical scan. rules are now sorted by id.

the second commit strips CR/LF from operator-supplied text (target,
module id) before it's written into a markdown heading line; an embedded
newline followed by "## " would otherwise render as an injected,
standalone heading.

### DIFF
--- a/internal/report/markdown.go
+++ b/internal/report/markdown.go
@@ -40,7 +40,7 @@ func Markdown(results []Result) []byte {
 	for i := 0; i < len(order); i++ {
 		target := order[i]
 		b.WriteString("## ")
-		b.WriteString(target)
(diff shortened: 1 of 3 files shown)


vmfunc (approved):
both real. sorting ruleIDs before building driver.rules kills the map-iteration nondeterminism, and stripping cr/lf in sanitizeHeading closes the `target\n## injected` heading break since target/module are scan input. comments match the code, tests hit the malicious case. in.

PR #338, opened 2026-07-09
fix(modules): search fhs/xdg paths for packaged installs

TBX3D:
built-in module discovery only looked next to the executable and in the
working directory, so a packaged install (module files under
/usr/share/sif or /usr/local/share/sif, following XDG_DATA_DIRS or the
freedesktop default) never found its bundled modules. resolveBuiltinDir
now probes those data dirs too, after the existing executable-relative
and working-dir candidates, so behavior for a dev checkout is unchanged.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -87,7 +87,9 @@ make
 
 sif looks for modules in these locations:
 
-- **built-in**: `modules/` directory next to the sif binary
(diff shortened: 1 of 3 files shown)


vmfunc (approved):
candidate ordering is right (exec-relative, cwd, then data dirs) and dev behavior is untouched. nice touch adding the IsDir() check in firstExistingDir, old code would've happily used a `modules` file. in.

PR #339, opened 2026-07-09
fix(git): validate .git exposure responses are real git artifacts

TBX3D:
a catch-all app that answers every path with 200 (a SPA shell, a custom
404 page) made every git-exposure path look present, since the check only
looked at the status code. responses are now validated against the actual
shape of the git artifact (blob/tree marker bytes, a sha1/sha256 object
name, a valid reflog line, ref file contents) before being reported, with
a capped read so a large catch-all body doesn't get buffered in full.

### DIFF
--- a/internal/scan/git.go
+++ b/internal/scan/git.go
@@ -15,7 +15,10 @@ package scan
 import (
 	"bufio"
 	"context"
+	"encoding/json"
+	"io"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
good fix. the catch-all 200 spa/soft-404 case was a real gap and status+content-type alone never covered it.

the things that make this correct:
- /logs/ matched before /refs/ so logs/refs/... hits the reflog branch, not the object-name branch. easy to get backwards.
- index checked against the raw body for DIRC instead of the trimmed text, right call since it's binary.
- the '[' -> json.Valid disambiguation for gitignore character classes ([Bb]in/) vs a json array shell. nice catch, and you tested it.

regexes are anchored with bounded char classes so no backtracking blowup, and the 8k LimitReader + DrainClose keeps the pool warm on a big shell body. symbolic refs like refs/remotes/origin/HEAD (ref: ...) would fall through the object-name check as a false negative, but that's marginal and the documented .gitignore plaintext soft-404 limit is fine, the structured artifacts carry the weight.

in.

PR #340, opened 2026-07-09
feat(fuzz): add fuzz harnesses for jwt, secrets and yaml parsing

TBX3D:
adds 7 fuzz targets across the parsers and extractors that read untrusted
input directly: YAML module definitions, JWT analysis and segment
decoding, endpoint/secret extraction from JS, OpenAPI spec parsing, and
<title> extraction from an HTTP body. all corpora are clean after a local
run (no crashers); each target asserts its own invariant beyond "must not
panic" where one exists (a reported secret match must be a verbatim
substring of the input, extracted endpoints must be sorted and deduped).

the YAML fuzz target needed parseYAMLModuleBytes pulled out of
ParseYAMLModule so module bytes can be fuzzed without a temp file; the
path-based function now just wraps it, no behavior change.


vmfunc (approved):
invariants line up with the code, which is the part i cared about. secretValue only ever returns a capture group or TrimSpace(groups[0]), both contiguous substrings, so the strings.Contains check in secrets_fuzz can't false-crash. same for endpoints: slices.Sort + isEndpoint filtering means sorted/non-empty holds. parseYAMLModuleBytes split is a clean no-op refactor. good seed spread on the none-alg / empty / malformed segment cases. in.

PR #341, opened 2026-07-09
fix(detectors): framework detector fp/fn corpus pass

TBX3D:
several framework detectors keyed on signatures that either dropped valid
signal or diluted a real match. django's csrfmiddlewaretoken pattern was
marked HeaderOnly even though it's a hidden form body field, so it never
matched. aspnet's weak .aspx/.ashx/.asmx path-extension signatures diluted
confidence below threshold on a response carrying only header markers;
they're dropped, and version extraction for aspnet/flask now also
searches canonical header text so header-shaped versions
(X-AspNet-Version, Werkzeug's Server header) are found.

a pinned fp/fn corpus test (hardening_test.go) locks in both the positive
and prose-false-positive case for each of the affected detectors.

### DIFF
--- a/internal/scan/frameworks/detectors/backend.go
+++ b/internal/scan/frameworks/detectors/backend.go
@@ -190,7 +190,9 @@ func (d *aspnetDetector) Detect(body string, headers http.Header) (float32, stri
 
 	var version string
 	if confidence > 0.5 {
-		version = fw.ExtractVersionOptimized(body, d.Name()).Version
+		// ASP.NET's strongest version signal is header-shaped
+		// (X-AspNet-Version), so search headers too, not just the body.
+		version = fw.ExtractVersionFromResponse(body, headers, d.Name()).Version
 	}
 	return confidence, version
 }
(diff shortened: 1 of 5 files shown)


vmfunc (approved):
the `(?i:X-AspNet-Version)` catch in version.go is the good one. http.Header canonicalizes to X-Aspnet-Version so the old case-sensitive literal could never have matched off a real response, and the comment spells out exactly why. django csrfmiddlewaretoken going HeaderOnly->body is right too, that's a hidden form field, never a header.

one thing, not blocking: bumping _next/static to 0.6 pushed the denominator to 1.7, so __NEXT_DATA__ alone is now 0.5/1.7 = 0.294, conf lands at 0.485 and it stops self-detecting (it used to at 0.333). in practice both markers co-occur double-quoted so real next pages still fire, but __NEXT_DATA__ is the most definitive next marker and you gave ng-version and astro self-clearing weight in this same pass, so it's odd to leave this one under the line. bump it to ~0.6 while you're in there.

rest is clean. in.

PR #343, opened 2026-07-09
fix(store,logger,scan): concurrency-adjacent races and correctness fixes

TBX3D:
a handful of fixes surfaced while working on -concurrency: two targets
whose sanitized names collided (differing only in separator punctuation)
could silently overwrite each other's state snapshot, so the snapshot
path now appends a hash of the full, un-sanitized target to stay
injective, and Save writes through a temp file + rename so a reader never
observes a partially-written snapshot. logger.CreateFile/Write now flatten
a URL's path into the filename instead of keeping the '/' verbatim, which
previously tried (and failed) to open a file under a directory that was
never created. a couple of scan/module status lines that bypassed the
output sink via a raw fmt.Println are routed through it, since only the
sink is lock-wrapped under -concurrency>1.

### DIFF
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -62,10 +62,9 @@ func (l *Logger) getWriter(path string) (*bufio.Writer, error) {
 		return w, nil
 	}
 
-	// belt-and-suspenders: buildLogPath already flattens '/' out of the
-	// filename so path's parent is always dir itself, but make sure that
-	// parent exists too in case a caller ever hands getWriter a nested path
-	// directly (e.g. a future caller, or a dir that was never Init'd).
+	// belt-and-suspenders: logPath already flattens '/' out of the filename
+	// so path's parent is always dir itself, but a future caller could still
+	// hand getWriter a nested path directly, so make sure it exists too.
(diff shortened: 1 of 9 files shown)


vmfunc (approved):
routing CreateFile and Write both through logPath is the right fix, that stray '/' was landing you in a never-created subdir and killing the target's scan. hash suffix on pathFor makes it injective, sanitize really was lossy on separator punctuation. atomic temp+rename is clean, same-dir so the rename actually stays atomic and a concurrent reader can't see a half-written snapshot.

two notes, neither blocking:

the "only the sink is lock-wrapped" line in the bypass tests and the pr body isn't right. sink is a bare os.Stdout, no mutex, each output.* is one fmt.Fprintf same as fmt.Println, so nothing's actually protecting against tearing there. the real win is output.Error/ScanStart respect apiMode and -silent, so next.go stops leaking raw errors straight into json/api stdout, which fmt.Println was bypassing. fix the comment so the reasoning matches what it's really doing.

the hash suffix orphans every existing snapshot on upgrade. first run after this re-flags everything as new and the old files never get matched again. fine as a one-time cost, just note it in the changelog.

in.

PR #344, opened 2026-07-09
fix(scan): tune js secret rules for false positives and missed formats

TBX3D:
two fixes to the js secret scanner: several rules had a trailing \b that
silently dropped a match whenever the key's last character (/, +, -) has
no word boundary there (aws secret keys, google api keys); and Stripe's
publishable pk_ keys and test-mode keys were being reported even though
neither is a live secret. the second commit adds detection for several
provider key formats that weren't covered at all (github fine-grained
PATs, gitlab PATs, anthropic keys, npm/google-oauth/stripe-webhook/
shopify/sendgrid tokens, slack incoming-webhook urls) plus a generic
apikey/secret/token="<value>" pattern gated on entropy and requiring a
digit, to avoid flagging camelCase identifiers.

### DIFF
--- a/internal/scan/js/secrets.go
+++ b/internal/scan/js/secrets.go
@@ -67,13 +67,12 @@ var secretRules = []struct {
 		minEntropy: noEntropyGate,
 	},
 	{
-		// github fine-grained personal access tokens: github_pat_ then 82 chars.
 		name:       "github fine-grained pat",
 		re:         regexp.MustCompile(`\b(github_pat_[0-9A-Za-z_]{82})\b`),
 		minEntropy: noEntropyGate,
 	},
 	{
-		// slack bot/user/app/legacy tokens plus version-anchored xapp app tokens.
+		// slack bot/user/app/legacy tokens, plus xapp tokens.
 		name:       "slack token",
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
solid, and the fp fixes are the meaty part: dropping the trailing \b on the aws-secret and google rules so keys ending in / + - stop getting silently dropped, cutting pk_ out of the stripe rule (publishable keys are public by design), and the requireDigit gate to kill camelCase generic matches. all correct, and the tests cover the real edges (slash/dash endings, spark_live boundary, wrong anthropic pad).

one thing before this lands: the second commit overlaps hard with #361. github_pat, npm, stripe webhook, sendgrid, slack webhook, gitlab and anthropic all show up in both, editing the same block of secretRules, so they'll conflict and double up coverage. decide which framing wins and drop the dupes from the other. these fp fixes are unique to this PR so i'd keep them here regardless. fwiw your anthropic rule (sk-ant-...{93}AA) is tighter than #361's {80,120}, keep this one.

minor, not blocking: you pulled the trailing \b off aws/google for exactly the '-'-ending case, but glpat-, GOCSPX- and the SG. rules still carry \b over a class that includes '-'. a key ending in '-' inside quotes gets missed the same way (google-oauth and sendgrid are fixed-count so total miss, gitlab {20,} just truncates the trailing dash). worth stripping \b there too.

in.

PR #345, opened 2026-07-09
fix(scan): capture every chunk in next.js build manifest arrays

TBX3D:
the pages-router build manifest regex anchored on the opening bracket of each route's chunk array, so it only ever captured the first chunk and silently dropped the rest whenever a route referenced more than one. tightened the regex to match every relative static/ chunk path instead of anchoring on the array boundary, without loosening it to a bare quoted-.js match, which would also pull in non-chunk strings such as __rewrites destinations (potentially an attacker-controlled absolute url). adds a realistic manifest fixture with shared IIFE-arg chunks and rewrite/sortedPages decoys to pin both the fix and the anti-fp behavior.


vmfunc (approved):
good catch on the anchor bug. per-chunk match is the right move, and requiring the `"static` prefix instead of a bare quoted-.js keeps the __rewrites absolute url out of the fetch list. the realistic fixture with the IIFE-arg shared chunks + decoys pins exactly that.

one loose end: the separator takes lowercase `/` via `[fF]`, but the ReplaceAll on line 74 only normalizes uppercase `/`, so a lowercase-escaped manifest would leave a literal `/` in the path. next emits uppercase so it's theoretical, fine to leave or just make the replace case-insensitive. in either way.

TBX3D:
rebased onto current main to clear the conflict. no content change to the fix
itself; the branch's tests needed updating to GetPagesRouterScripts' timeout
parameter, which main grew after this branch was cut.

PR #346, opened 2026-07-09
fix(modules): guard regex extractor against negative group index

TBX3D:
a module's extractor group is a plain int sourced straight from its yaml definition, so a module setting group: -1 hit matches[-1] and panicked at runtime. skip negative groups instead of indexing into the match slice.


vmfunc (approved):
yep, len(matches) > -1 is always true so matches[-1] was reachable from any module yaml. guard's right, test covers it. in.

PR #347, opened 2026-07-09
feat(modules): add crlf and ssi injection detection modules

TBX3D:
adds detection modules for crlf (header/response splitting) and ssi injection. both matchers are line-anchored and scoped to avoid the obvious false positives: the crlf regex requires the payload marker at the start of a header/body line rather than a bare substring match, and the ssi matcher excludes numeric-entity and raw-percent-encoded reflections that a naive regex would otherwise flag. test coverage includes adversarial handlers that reflect the payload in decoded and encoded forms without actually being vulnerable.

### DIFF
--- a/internal/modules/crlf_ssi_injection_test.go
+++ b/internal/modules/crlf_ssi_injection_test.go
@@ -1,0 +1,198 @@
+package modules_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"regexp"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/vmfunc/sif/internal/modules"
+)
+
+const (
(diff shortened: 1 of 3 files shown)


vmfunc (approved):
the fp reasoning holds up. the crlf `(?m)^X-Sif-Injected:` leans on getPart rendering each parsed header as its own `Key: value` line, so a real split lands at line-start while a mid-value reflection can't, and the ssi `[^<&%]*\d[^<&%]*` cleanly rejects literal `<`, entity `&` and raw `%3C` reflection since the directive carries no digit of its own. re2 backend so the two `[^<&%]*` runs aren't a redos risk. handlers (ssiEntityHandler, ssiRawHandler, the header-value reflection case) pin exactly the cases a naive matcher would trip on. in.

PR #348, opened 2026-07-09
fix(scan): decode supabase jwt body as base64url

TBX3D:
the supabase anon-key jwt body was being decoded with base64.RawStdEncoding, which rejects the url-safe - and _ characters that base64url actually uses. switched to a raw-url-encoding decode with a padded fallback for tokens that weren't stripped of padding, and added a nil check since a literal json null unmarshals into a nil pointer without an error.


vmfunc (approved):
the base64url swap is the right call. RawStdEncoding was silently dropping any payload with a 62/63 group (- or _), so real anon tokens were getting skipped as false negatives. the json-null guard is a good catch too, old code would've nil-derefed on supabaseJwt.ProjectId if a middle segment decoded to a literal null.

one cosmetic thing: the guard checks len(parts) < 2 but the message says 'want 3', and the 'too few segments' case (hdr.sig) actually errors at the json decode rather than the length check. not blocking, the regex guarantees 3 segments upstream anyway.

in.

PR #349, opened 2026-07-09
fix(scan): match git html filter case-insensitively

TBX3D:
the exposed-.git check compared the response Content-Type against text/html verbatim, so a server sending Text/HTML (or any other casing) for a soft-404 slipped past the filter and got reported as an exposed git directory. lowercase the content-type before the prefix check. added a regression test for the mixed-case case and a companion test asserting non-html content types still fire, so the fix doesn't over-correct into suppressing real hits.

### DIFF
--- a/internal/scan/git.go
+++ b/internal/scan/git.go
@@ -88,7 +88,9 @@ func Git(url string, timeout time.Duration, threads int, logdir string) ([]strin
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


TBX3D:
closing as superseded by #339, which replaces this whole content-type check with a body-content validator that doesn't rely on content-type at all.

PR #350, opened 2026-07-09
fix(scan): resolve favicon href against origin to stop pathful-target misfetch

TBX3D:
a root-relative <link rel=icon> href was being appended to the target url's existing path instead of resolved against the origin, so a target like https://example.com/app would fetch /app/favicon.ico instead of /favicon.ico. resolveFaviconURL now delegates to net/url's ResolveReference, matching how a browser actually resolves root-relative, scheme-relative and absolute icon hrefs.

### DIFF
--- a/internal/scan/favicon.go
+++ b/internal/scan/favicon.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
swapping the prefix ladder for ResolveReference is the right call, and routing the /favicon.ico probe (favicon.go:136) through it too is what actually kills the /app/favicon.ico misfetch, not just the declared-link case. base always carries a scheme post-normalize so the parse-error fallbacks are dead but harmless. anchoring tests cover all five branches. in.

PR #351, opened 2026-07-09
fix(scan): fetch wayback source over https

TBX3D:
the wayback machine passive-recon feed was being fetched over plain http, so an on-path attacker able to tamper with the response could inject or drop subdomains/urls from what's meant to be a trusted enumeration source. switched the feed's base url to https, matching the other two passive sources (crt.sh, certspotter), and added a test pinning all three feeds to https.


vmfunc (approved):
yeah, web.archive.org serves the cdx endpoint over tls, no reason we were on plain http. good that the test pins all three feeds and not just wayback. in.

PR #352, opened 2026-07-09
feat(pool): add EachCtx for cancellable fan-out

TBX3D:
Each has no way to stop early: once -max-time expires or ctrl-c fires, in-flight workers keep draining the whole queue before the pool returns. EachCtx takes a context and stops handing out new items once it's cancelled, while workers already running finish their current item rather than being killed mid-write. Each is now a thin wrapper over EachCtx with context.Background().

### DIFF
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -16,12 +16,26 @@
 // holding it, the rest keep draining the queue instead of idling behind it.
 package pool
 
-import "sync"
+import (
+	"context"
+	"sync"
+)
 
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
ctx.Err() check sits right before fn so a cancel stops new pulls but never kills a worker mid-item, which is the behavior you want on -max-time. defer wg.Done() on every exit path so the early return can't leak, and Each collapsing to the Background() wrapper is a no-op on the old path. in.

PR #353, opened 2026-07-09
fix(output): truncate progress item on a rune boundary

TBX3D:
the progress line truncated long items with a byte-index slice, which can land inside a multibyte rune (an idn host, a unicode route) and render as a garbled replacement glyph. truncateItem now counts and cuts on runes instead. bundled with a second, unrelated notify fix: slack/discord finding text is untrusted and could carry its own triple-backtick run or channel/user mention entities, so it's neutralized before being fenced.

### DIFF
--- a/internal/notify/notify_test.go
+++ b/internal/notify/notify_test.go
@@ -159,6 +159,41 @@ func TestWebhookPayloadShape(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


TBX3D:
closing as superseded by #336: the only functional change here (rune-safe progress-bar truncation) is already covered there, and the other file touched in this commit (slack.go) is a non-functional comment reshuffle with nothing left to keep once the duplicate truncation fix is removed.

PR #354, opened 2026-07-09
fix(scan): decode html entities in probe title

TBX3D:
the probe's extracted page title carried raw html entities (Tom &amp;amp; Jerry) instead of the decoded text a browser would render, since there was no entity-decoding step after pulling the <title> text out. decode entities so the title matches the rendered page.

### DIFF
--- a/internal/scan/probe.go
+++ b/internal/scan/probe.go
@@ -15,6 +15,7 @@ package scan
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
unescape-before-trim is the right order here, an edge entity that decodes to whitespace still gets trimmed. title's display-only so nothing downstream regresses. in.

PR #355, opened 2026-07-09
refactor: share path-resolution and body-cap helpers

TBX3D:
the per-user sif config directory resolution (~/.config/sif on unix, %LOCALAPPDATA% on windows) and the 5MiB response-body cap were each copy-pasted across module loading and framework detection. pulled the first into internal/sifpath.UserSubdir and the second into httpx.ReadCappedBody, with the same resolved paths and cap as before, verified by new tests on the path-layout invariants.

### DIFF
--- a/internal/httpx/httpx.go
+++ b/internal/httpx/httpx.go
@@ -68,6 +68,11 @@ const (
 // conn, so we cap the read and let the conn be discarded instead.
 const drainCap = 16 << 10
 
+// MaxBodySize is the shared ceiling on how much of a response body the scanners
+// read into memory, so a hostile or accidental multi-gigabyte response can't
+// exhaust the process.
+const MaxBodySize = 5 << 20
+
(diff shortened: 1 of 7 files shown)


vmfunc (approved):
solid. sifpath.UserSubdir preserves the exact layout for both modules and signatures, 5 << 20 matches the drainCap style in httpx.go and equals the old 5*1024*1024, and ReadCappedBody leaving the close to the caller is fine since both callsites still defer Body.Close(). in.

one day we should sweep the remaining inline 5*1024*1024 in cms/patchnotes/supabase/etc onto MaxBodySize too, but that's out of scope here.

TBX3D:
rebased onto current main to clear the conflict. the only real overlap was
loader.go, where main has since grown resolveBuiltinDir (executable dir, working
dir, then the XDG data dirs) plus the embedded module FS. resolved in favour of
main's version, with sifpath owning only the per-user lookup, which is what this
branch was actually extracting.

PR #356, opened 2026-07-09
feat(modules): add fingerprint module type

TBX3D:
a fingerprint module identifies a technology by weighted body/header signatures scored into a confidence, with an optional version regex, rather than a boolean match. it fires one finding carrying the score once it reaches the threshold (default 0.5) - the framework detectors' scoring approach, in the module format, so a custom tech fingerprint can live alongside other modules. validated at load (signatures present, non-empty patterns, finite weights, confidence in [0,1], version regex compiles) and covered by yaml round-trip, validation, header, version and default-threshold tests. documented in docs/modules.md.

depends on #355: the first two commits here are that PR's shared sifpath/httpx helpers (fingerprint.go uses httpx.ReadCappedBody), included so this branch builds standalone. once #355 merges, rebasing this one should drop those two commits as already-applied and leave only the fingerprint commit.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -65,7 +65,8 @@ info:
 
 ### type (required)
 
-module type. `http` and `tcp` are supported.
+module type. `http` (request and match), `tcp` (raw connection probe) and
+`fingerprint` (weighted technology detection) are supported.
 
 ```yaml
 type: http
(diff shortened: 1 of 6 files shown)


vmfunc (approved):
scoring lines up with the framework detector's MatchSignatures, and the boundary's right: score == threshold fires, 0.4 body-alone stays quiet. validation covers the nan/inf weight and regex-compile cases too, good.

two non-blocking notes:
- version + evidence are both body-only. scoreFingerprint runs the regex on body, and the finding's evidence is truncateEvidence(bodyStr). so a header-scoped fingerprint like your acme example can't pull a `Server: nginx/1.2.3` version, and the evidence won't contain the thing that actually matched. worth a follow-up, a lot of real fingerprints are server-header versions.
- no user-agent on the probe, the http executor sets `sif/1.0`. some hosts vary or block on go's default UA, so this can move detection accuracy. cheap to add.

confidence: 0 quietly becomes 0.5 via the `== 0` check so you can't express always-fire, minor corner. in.

TBX3D:
rebased. #355 merged, so the two shared-helper commits this was stacked on dropped as already-upstream and the branch is just the fingerprint module type now. the only resolution was the docs `type` line, which lists http, tcp and fingerprint.

PR #357, opened 2026-07-09
feat(config): add -config/-profile config file and profile overlays

TBX3D:
adds a -config flag to load flag defaults from a yaml file, and a -profile flag to select a named overlay within it, so a recurring scan setup doesn't need to be retyped as flags every run. explicit cli flags still win over anything in the config/profile - a flag's DefValue is compared against what the user actually passed so an unset flag can be overridden by the file without a set flag being silently clobbered. includes malformed-config validation on the no-profile path and a documented example config.


vmfunc (approved):
solid. the load-bearing bit is the explicit-key strip in buildFlatConfig: goflags treating flag==DefValue as unset would silently let the file win over an explicit `-timeout 10s`, and pulling those keys out of the map before it ever reaches goflags is the right fix. good that it's pinned by TestExplicitFlagAtDefaultBeatsFile and the short-alias variant, that's exactly the case that'd rot quietly otherwise.

flagAliasGroups deriving the alias table from registerFlags instead of a second hardcoded name map is the right call, it can't drift.

and routing the no-profile path through loadConfigMap so a malformed -config errors loudly instead of goflags swallowing the decode and running with every setting dropped, that's a real bug you closed, not just tidiness. TestMalformedConfigNoProfileErrors covers it.

one sharp edge, non-blocking: explicitFlagTokens counts any `-x` arg as a flag token, so a string-flag value that's literally another flag name (`-cookie -t`, cookie value "-t") would strip `timeout` out of the config and revert it to the built-in default. contrived enough that i wouldn't hold the PR for it, and fixing it properly means reintroducing the arity table you deliberately dodged. leave it, maybe a comment noting the limitation.

ship it.

PR #358, opened 2026-07-09
refactor(fingerprint): make the favicon tech table the single source

TBX3D:
the shodan favicon-hash to tech-name table lived in internal/scan/favicon.go with a separate demo-sync test keeping it lined up against the favicon-*.yaml demo modules. moved it into internal/fingerprint as LookupFaviconTech so there's one table instead of two things that have to agree, and wired the module-side favicon evidence string to name the matched tech (favicon mmh3=<n> tech=<name>) instead of a bare hash.


vmfunc (approved):
map moves over byte-identical and the demo drift-guard test moves with it instead of getting dropped, so the two-table-agreement risk actually goes away, which is the whole point. `tech, _ := LookupFaviconTech(hash)` in scan/favicon.go discarding the ok is fine, "" is the documented miss value so it's behavior-equal to the old `faviconHashes[hash]`, and it's a bool not an error. pinning the count at 10 in TestLookupFaviconTech is the right guard. in.

PR #359, opened 2026-07-09
feat(modules): bridge fingerprint modules into framework detection

TBX3D:
a fingerprint module and a framework detector can independently identify the same technology. this registers default-confidence, root-scoped fingerprint modules into the framework detector registry so framework detection covers yaml-defined technologies, and keeps the two engines from naming the same thing twice.

BridgeFingerprints is called from Run before scanAllTargets. it has to sit there: the per-target framework scan runs inside the -concurrency worker pool, so bridging any later would not reach it, and loading up front keeps the pool off the shared registry write path. the module set is now loaded once per run rather than once per target.

BridgeFingerprints returns the ids it bridged, and an implicit module run (-all-modules, -module-tags) skips them, so the technology is reported by the framework engine alone. naming a module with -modules is a direct request and still runs it. the two engines disagree on one boundary, the module engine fires at score >= confidence and the framework engine at score > threshold, so a bridged module sitting exactly at 0.5 is reported by neither in an implicit run. that is the residual bridgeableToFramework documents.

the bridged name is the module id and frameworks.Register clobbers by name, so a module called "WordPress" would have replaced the builtin detector. frameworks.RegisterIfAbsent does the check and the insert under one lock and the bridge refuses the name instead.

also marks the legacy custom-signature (signatures dir) path as deprecated in favour of fingerprint modules, without removing it.

stacked on #355 (shared sifpath/httpx helpers) and #356 (fingerprint module type), both open. this branch includes their commits so it builds standalone; rebasing after they merge should drop those as already-applied.

### DIFF
--- a/internal/modules/bridge.go
+++ b/internal/modules/bridge.go
@@ -1,0 +1,123 @@
+/*
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+:                                                                               :
+:   █▀ █ █▀▀   ·   Blazing-fast pentesting suite                                :
+:   ▄█ █ █▀    ·   BSD 3-Clause License                                         :
+:                                                                               :
+:   (c) 2022-2026 vmfunc, xyzeva,                                               :
+:                 lunchcat alumni & contributors                                :
+:                                                                               :
+·━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━·
+*/
+
+package modules
+
+import (
+	"net/http"
+	"regexp"
+
+	"github.com/charmbracelet/log"
+	"github.com/vmfunc/sif/internal/scan/frameworks"
+)
+
+// bridgeableToFramework reports whether def is a fingerprint whose semantics
+// the framework engine can reproduce exactly: root path, default confidence,
+// all weights > 0. anything else stays module-only. the two engines still
+// disagree at score == 0.5 (>= confidence vs > detectionThreshold).
+func bridgeableToFramework(def *YAMLModule) (bool, string) {
+	if def.Type != TypeFingerprint || def.Fingerprint == nil {
+		return false, "not a fingerprint module"
+	}
+	cfg := def.Fingerprint
+	if cfg.Path != "" && cfg.Path != "/" {
+		return false, "non-root path"
+	}
+	if cfg.Confidence != 0 {
+		return false, "custom confidence"
+	}
+	for _, s := range cfg.Signatures {
+		if s.Weight <= 0 {
+			return false, "zero-weight signature"
+		}
+	}
+	return true, ""
+}
+
+// bridgeFingerprint registers def as a framework detector when
+// bridgeableToFramework allows it, and reports whether it registered. a guard
+// failure is not an error: the module still runs natively in the module
+// engine, so bridging is pure upside or a no-op.
+func bridgeFingerprint(def *YAMLModule) (bool, string) {
+	ok, reason := bridgeableToFramework(def)
+	if !ok {
+		return false, reason
+	}
+
+	cfg := def.Fingerprint
+	sigs := make([]frameworks.Signature, len(cfg.Signatures))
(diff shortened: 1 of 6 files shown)


vmfunc (changes requested):
the fingerprint module type itself is solid. scorer parity with MatchSignatures is pinned properly, and the zero-weight remap divergence + strict-threshold-at-0.5 edges are exactly right.

but the bridge doesn't run. BridgeFingerprints has no caller, grep it, only the definition and the tests reference it. at runtime nothing ever lands in the framework registry, so as landed this is inert scaffolding, not what the title says.

and even once you wire it, it doesn't dedup. the fingerprint module still executes in the module engine and reports its own finding (your bridge comment says as much), and registering it as a framework detector just adds a second surface, the framework engine already collapses to one best. so the tech shows up in both the module output and the framework output regardless. bridging never removes the native report. if the plan is to drop bridged fingerprint modules from the module run, that logic isn't in here.

so before this goes in: is the wiring a follow-up, and where does the native finding get suppressed? one of those has to exist for this to do what it claims. and fix the description, it states the dedup as done.

minor: bridged name is def.ID and frameworks.Register clobbers by name, so a module id colliding with a builtin ("WordPress", "Django"...) silently overwrites the builtin detector. skip-or-warn on an existing key.

TBX3D:
you are right on both counts and the description was the worst of it, sorry, it
described a dedup that nothing performed.

wiring: BridgeFingerprints is now called from Run before scanAllTargets. it has
to be there rather than anywhere later, because the per-target framework scan
runs inside the -concurrency worker pool, so bridging after that point never
reaches the detection it is meant to feed. that also meant loading the module set
once per run instead of once per target, which the pool wanted anyway.

dedup: BridgeFingerprints returns the ids it bridged and an implicit module run
(-all-modules, -module-tags) skips them, so the framework engine reports the
technology alone. naming a module with -modules is a direct request and still
runs it, which felt less surprising than silently dropping something explicitly
asked for. the 0.5 boundary you would expect from that: the module engine fires
at score >= confidence and the framework engine at score > threshold, so a
bridged module sitting exactly at 0.5 is reported by neither in an implicit run.
that is the same residual bridgeableToFramework already documents, and it is
called out in the comment at the call site. happy to close it instead by
suppressing at report time if you would rather not carry it.

collision: added frameworks.RegisterIfAbsent, check and insert under one lock,
and the bridge refuses a taken name rather than replacing it. covered by a test
that registers a detector first and asserts the original still matches its own
marker afterwards.

PR description rewritten to describe what the code does.

TBX3D:
follow-up on a regression I introduced in the wiring commit.

folding the two module-load sites into one helper made loadModules swallow a
loader construction failure and warn, but -list-modules used to treat that as
fatal. it would have printed an empty module list and exited 0.

loadModules now returns an error: -list-modules propagates it, a scan still
downgrades to a warning, which is what each path did before they shared the
helper.

also stopped the bridged-module filter from reusing the input slice's backing
array; it builds a new slice now, so it cannot disturb a caller's view of the
registry if All()/ByTag() ever hand back shared storage.

TBX3D:
rebased onto current main now that #355 has landed. the two shared-helper
commits dropped as already-upstream, as the description predicted, so this is
down to the 8 commits that are actually the fingerprint type plus the bridge.

heads up that CI on this branch (and a few others) is currently red for a reason
unrelated to the change: main does not build at 2ce359a8,
internal/modules/executor.go:197 references MaxBodySize after #355 moved it to
internal/httpx. #374 fixes it. verified this branch builds and tests clean with
that one-line fix applied on top.

vmfunc:
this is the good half of the pair - the bridge reasoning and the RegisterIfAbsent guard are exactly right. #356 just merged so rebase this down to just bridge.go + the wiring (no sense landing the fingerprint type twice) and i'll take it.

---

TBX3D:
done — rebased onto current main (post-#356) and reduced to the two commits you asked for:

- `feat(modules): bridge root fingerprints into framework detectors` — bridge.go + the yaml.go helper + bridge tests
- `feat(cli): wire fingerprint bridging into the run and stop double-reporting` — the wiring in sif.go, RegisterIfAbsent, and the dedup test

dropped everything that belonged to the fingerprint type or was out of scope: the signatures-dir deprecation notice, the scorer-equivalence tests, and the legacy custom-signature backcompat test. no more overlap with #356. green on fmt/vet/build/test/race/golangci-lint.

PR #360, opened 2026-07-09
feat(frameworks): add web application firewall detectors

TBX3D:
adds detectors for cloudflare's bot-challenge/mitigation response, sucuri, imperva incapsula, f5 big-ip asm, and modsecurity. each keys on a header, set-cookie name, or block-page string the waf itself sets on an interception response, rather than anything that could appear in ordinary prose about the vendor, so a page merely mentioning a waf by name doesn't trip it.


vmfunc (approved):
the normalization is doing the right thing on the proxy markers, x-sucuri-cache and cf-ray both sit under the bar on their own so a plain sucuri/cf proxy doesn't get read as a waf block. that's the part that usually goes wrong, good.

one thing on waf.go, the f5 asm detector: "Your support ID is:" is weight 0.3 against a 0.9 total, so it normalizes to 0.33 score and clears 0.5 confidence by itself, no rejection sentence needed. your comment says it's block-page driven off the reject line, so i don't think you meant the support-id line to stand alone. as-is it's a soft fp on any support/helpdesk page that uses that phrasing. either knock its weight down so it can't clear on its own, or gate it behind the reject sentence.

not a blocker, land it and tweak after if you want. in.

PR #361, opened 2026-07-09
feat(js): add provider-prefixed and structural secret rules

TBX3D:
extends the secret bank with eleven providers it does not cover: pypi, openai (legacy and project/service-account), square, mailgun, discord (bot tokens and incoming webhooks), new relic, cloudinary, jwts, and `scheme://user:pass@host` connection strings.

the last two have no fixed prefix, so shape alone is not proof. a `validate` hook runs after the entropy and digit gates: the jwt rule decodes the header segment and requires `alg` and `typ`, which drops arbitrary dot-separated base64url blobs, and the connection-string rule rejects a userinfo password drawn from the usual documentation placeholders.

originally this added nineteen rules. main has since grown its own bank covering gitlab, anthropic, npm, sendgrid, slack webhooks, github fine-grained pats, google oauth, shopify and stripe (whose `(?:sk|rk)_live_` shape already covers restricted keys), so every rule that would have duplicated one of those is dropped rather than shipped a second time under a different name. that also settles the #344 overlap noted in the original description.

`TestProviderRulesDoNotDuplicateExistingCoverage` pins that no credential is reported twice by two prefixed rules. it excludes the generic assignment rule on purpose: that rule claims any quoted high-entropy value behind a token/password/secret keyword, so it already doubles up with every prefixed rule on main. that is pre-existing and out of scope here, but worth flagging.

### DIFF
--- a/internal/scan/js/secrets.go
+++ b/internal/scan/js/secrets.go
@@ -13,6 +13,7 @@
 package js
 
 import (
+	"encoding/base64"
 	"math"
 	"regexp"
 	"strings"
@@ -39,12 +40,14 @@ const (
 
 // secretRules is the credential regex bank. the matching group (or the whole
 // match when there's no group) is what gets reported; minEntropy gates the
-// generic high-entropy rules so we don't flag every short literal.
+// generic high-entropy rules so we don't flag every short literal. validate
+// runs after the gates for rules where shape alone isn't proof.
 var secretRules = []struct {
 	name         string
 	re           *regexp.Regexp
 	minEntropy   float64
 	requireDigit bool
+	validate     func(string) bool
 }{
 	{
 		// aws access key ids are fixed-shape and unmistakable.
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
the validate hook is the right shape here. gating jwt and the conn-string on actual structure instead of widening the regex keeps the fp floor where it belongs, and the placeholder-password filter will kill a ton of .env.example noise.

two small things, neither blocking: isStructuredJWT requires "typ" but that's optional per 7515/7519, plenty of real id tokens omit it so you'll miss some. alg alone is the mandatory one. and key-[0-9a-f]{32} for mailgun trips on any cache/redis key named key-<md5>, it's the one loose rule in the set. fine to land and tighten later.

the #344 restricted-key / fine-grained-pat overlap is on you two to reconcile whenever the second one lands. in.

TBX3D:
rebased, and this one needed re-scoping rather than a replay. main grew its own rule bank in the meantime, covering gitlab, anthropic, npm, sendgrid, slack webhooks, github fine-grained pats, google oauth, shopify and stripe.

a straight rebase left eight rules duplicated by name and three more duplicated by shape: main's stripe rule is `(?:sk|rk)_live_`, so the separate restricted-key rule here had nothing left to add, and "gitlab access token" was main's "gitlab personal access token" regex under a different name. shipping that would have double-reported every one of those credentials.

so this is down to the eleven providers main does not cover, plus the two rules that need validation rather than a shape (jwt, connection strings). that also settles the #344 overlap i flagged in the original description. `TestProviderRulesDoNotDuplicateExistingCoverage` pins that no credential is claimed by two prefixed rules.

one thing this does not fix, surfacing it rather than quietly leaving it: the generic assignment rule claims any quoted high-entropy value behind a token/password/secret keyword, so `const token = "<jwt>"` is reported twice, once by the prefixed rule and once by the generic one. that is already true on main for every prefixed rule, so i left it out of scope here, but happy to take it separately if you want it closed.

PR #362, opened 2026-07-09
feat(modules): add a range matcher for response size or status

TBX3D:
the existing size matcher only checks exact-match against a list of values. this adds a range matcher type with inclusive min/max bounds over either the response/banner byte length (source: size, the default) or the http status code (source: status). an unbounded range, an inverted one (min > max), or an unknown source is rejected at load rather than silently never firing at match time.


vmfunc (approved):
validation covering the three ways a range silently never fires (unbounded, inverted, unknown source) is the right call, and inRange's nil-bound handling matches the docstring exactly. the favicon validateMatchers restructure preserves the old behavior 1:1. tests hit the inclusive boundaries and the reject paths, good.

one gap in the same spirit as the rest: source:size measures len(body), which is capped at MaxBodySize (5mb) by the LimitReader. so a min above 5mb passes load-time validation but never fires, the exact thing you're guarding against everywhere else. not a blocker, worth a clamp or a note on the Min/Max field doc. in.

PR #363, opened 2026-07-10
feat(frameworks): add detectors for 56 additional cms, frontend and backend platforms

TBX3D:
adds detection for a batch of previously-uncovered platforms: cms (bitrix, blogger, bolt, concrete, contao, craft, discourse, dotnetnuke, expressionengine, flarum, grav, hubspot, liferay, mediawiki, moodle, nodebb, october, opencart, plone, prestashop, sitecore, squarespace, statamic, textpattern, typo3, webflow, wix, woocommerce, xenforo, xwiki), static-site generators (hugo, jekyll, docusaurus, mkdocs, eleventy, hexo, vuepress, sphinx, nikola, publii), frontend libs (alpine.js, knockout, livewire, qwik, stimulus, turbo, unpoly) and backend frameworks (tornado, cherrypy, play, sails.js, beego, jsf, gwt, vaadin, coldfusion). also tightens false-positive guards on 5 existing detectors (adonisjs, cakephp, codeigniter, phoenix, strapi).

the hosting/edge detectors and jquery are no longer part of this batch. every detector registers into one pool and DetectFramework reports a single argmax across it, so a single-signature edge marker scores 1.0 and sigmoids to the ceiling while a multi-signature framework that lands only its primary signal scores lower. a wordpress page behind cloudflare resolved to "Cloudflare" at 0.9991, and a django app shipping jquery resolved to "jQuery" at 0.9844. both are pinned by argmax_shadowing_test.go so the batch cannot regrow the problem.

the cdn side belongs in the separate DetectCDN pool (#287); jquery wants a surface that can report a technology without competing for the single framework slot.


vmfunc (changes requested):
the cms/frontend/backend/meta detectors are solid and the fp tightening is the good part here: `_sf2_` over the bare `sf_` substring (backend.go, with the misfa_sf_token near-miss actually tested), strapi scoped to x-powered-by instead of a body "strapi" match, cakephp dropping the bare body match down to just the CAKEPHP cookie, and the stimulus data-controller/data-action split so neither clears the bar alone. the negatives exercise the real collisions (statamic_ vs statamic.com, data-turbo- vs data-turbolinks, akamai vs akamaihd.net). no fp bug in any of those.

hosting.go is the blocker though. every detector registers into one pool and DetectFramework (detect.go:106) reports a single best across the whole registry. the cdn/host detectors are single-signature edge-header matches, so on a hit they score weightedScore/totalWeight = 1.0 -> sigmoid ~0.999, the ceiling. a multi-sig framework that only lands its primary signal lands lower (wordpress with wp-content+wp-includes+wp-json ~0.98). so any real target behind cloudflare/cloudfront/vercel/netlify now reports the cdn as its "framework" and the actual app framework gets shadowed. that's the common case, not an edge.

drop hosting.go from this batch and let it land as the dedicated cdn module in #287, which is the right home and which you already flagged. or if it has to live here, DetectFramework needs to surface host/cdn as a separate axis rather than compete for the single framework slot. the other three files are good to go once hosting is out.

smaller version of the same thing: jquery is on most of the web, so 'jquery-'/'jquery.min.js' matching at ~0.6+ will start winning over real frameworks on plenty of sites too. not blocking, but same shadowing root cause, worth folding into whatever you do for the selector.

TBX3D:
reproduced it before touching anything. a wordpress page (wp-content,
wp-includes, wp-json) served behind cloudflare:

  winner="Cloudflare" confidence=0.9991

dropped hosting.go and hosting_test.go from the batch; the cdn side belongs in
the DetectCDN pool in #287. same test now gives WordPress at 0.9797.

took the jquery note the same way rather than tuning it, because weights cannot
fix it: MatchSignatures normalizes by total weight, so only the ratios matter,
and a single "jquery.min.js" reference alone scores 0.5/1.4 -> 0.639, already
over the floor. a django app shipping jquery gave:

  winner="jQuery" confidence=0.9844

so jquery came out too, and it is Django at 0.7982 now. both cases are pinned in
argmax_shadowing_test.go so the batch cannot regrow the problem. if you would
rather jquery ship now and ride the multi-report work in #314 instead, say the
word and I will put it back, it is a one-commit revert.

the other four files are untouched. also folded in a duplicate hdr test helper
that collided with the one main grew in hardening_test.go.

PR #364, opened 2026-07-10
fix(scoring): clamp fraction scorers to [0, 1]

TBX3D:
a fraction scorer could return a value outside 0-1 depending on input distribution, which then propagated into downstream weighted sums. clamp at the source instead of trusting every caller to bound it.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -65,7 +65,8 @@ info:
 
 ### type (required)
 
-module type. `http` and `tcp` are supported.
+module type. `http` (request and match), `tcp` (raw connection probe) and
+`fingerprint` (weighted technology detection) are supported.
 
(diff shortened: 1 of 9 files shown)


vmfunc (approved):
test

TBX3D:
rebased onto the updated #356. no content change.

vmfunc:
the clamp is exactly right - nan guard, no-op on already-valid scores, good tests. but the diff isn't a scoring fix, it's the whole fingerprint type, because it was stacked on #356 (just merged). rebase onto main and it shrinks to just the clamp + MatchSignatures bit - one-click merge. also clampUnit is copy-pasted across two packages, pull it into a shared helper before it spreads.

---

PR #365, opened 2026-07-10
fix(dns): enforce a timeout floor alongside the resolver pool

TBX3D:
the resolver pool change let a caller-supplied timeout of zero (or unset) slip through uncapped, so a hung resolver could block a scan indefinitely. enforce a minimum floor regardless of what the caller passes.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -65,7 +65,7 @@ info:
 
 ### type (required)
 
-module type. `http` and `tcp` are supported.
+module type. `http`, `tcp` and `dns` are supported.
 
 ```yaml
 type: http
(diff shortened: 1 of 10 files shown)


vmfunc (approved):
solid. the RawResp-over-Raw choice in newDNSResponse (dns.go:249) is the right call, data.Raw concatenates across the retry loop so a matcher on it would double-count. and flooring only a non-positive timeout plus the black-hole listener test that pins the actual bound is exactly the shape this needed.

two small things, neither blocking:
- dns-spf-record.yaml:34, spf_qualifier `([-~?+]all)` matches the first `[-~?+]all` anywhere in the joined answer, so an include host like mail-all.example.com would mis-report the policy. anchor it to the trailing qualifier.
- validateDNS (dns.go:181) rejects a bad matcher type at load but lets a non-regex extractor through to silently no-op at runtime. docs say extractors are regex-only on dns, so reject it at load too for parity.

conflicting with main, needs a rebase before it goes in.

TBX3D:
rebased onto the updated #307. no content change of its own; dns.go picks up the `checkWords` case-insensitivity argument from that branch's conflict resolution.

vmfunc:
the timeout floor is right and the black-hole test proving it actually bounds is lovely. problem is the diff - this is stacked on #307, so what's really here is the whole dns module + a go.mod dep promotion, not a one-line floor. get #307 in first, rebase this on top, and it becomes the small fix it says it is. holding til then.

---

PR #366, opened 2026-07-10
fix(tcp): stop a cancelled context from being swallowed by the deadline re-arm

TBX3D:
ExecuteTCPModule cancels a run by tripping the connection with a single SetDeadline(now) from its watchdog goroutine. arming a later deadline after that trip silently pushes it back out to the full timeout, so a cancelled scan kept blocking: readTCP called SetReadDeadline unconditionally, and the probe write armed SetWriteDeadline the same way.

check ctx before arming either deadline and again on each read iteration, and prefer ctx.Err() over the raw write error so a write the watchdog trips is reported as a cancellation rather than an i/o timeout.

rebased onto main, which already carries the executor from #308, so this is now just the fix and its tests. TestExecuteTCPModuleContextCancelDuringWrite drives the write-in-flight ordering rather than asserting it in the abstract.

### DIFF
--- a/internal/modules/tcp.go
+++ b/internal/modules/tcp.go
@@ -101,14 +101,27 @@ func ExecuteTCPModule(ctx context.Context, target string, def *YAMLModule, opts 
 	}()
 
 	if cfg.Data != "" {
+		// same re-arm hazard readTCP guards against: if the cancel watchdog's
+		// SetDeadline(now) trip lands before this SetWriteDeadline call, the
+		// call below would silently push the deadline back out to the full
+		// timeout. Check ctx first so an already-cancelled run never arms it,
+		// and prefer ctx.Err() over the raw write error so a write the
+		// watchdog does trip is reported as a cancellation, not an i/o
+		// timeout.
+		if err := ctx.Err(); err != nil {
+			return result, err
+		}
 		payload := decodeTCPData(cfg.Data)
 		_ = conn.SetWriteDeadline(time.Now().Add(timeout))
 		if _, err := conn.Write([]byte(payload)); err != nil {
+			if ctxErr := ctx.Err(); ctxErr != nil {
+				return result, ctxErr
+			}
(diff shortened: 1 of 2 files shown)


vmfunc (approved):
the cancel-during-write re-arm race is the subtle part here and you handled it right: ctx.Err() before SetWriteDeadline and again at the top of readTCP, with TestExecuteTCPModuleContextCancelDuringWrite actually reproducing it instead of asserting it in the abstract. nice.

one non-blocking thing worth a follow-up: readTCP drains to the read deadline, so any peer that banners and then holds the connection open (ssh, redis after INFO, the mysql handshake, memcached after version... basically all of these) pays the full timeout on every grab. strictly better than the old hang-forever, but at the 10s default that's 10s per module on the happy path. the tests don't surface it because they all close the server side (the real listener defers c.Close, the pipe tests server.Close), so you get EOF instead of the hold-open you hit in the wild. an early return once a banner lands and the read goes idle, or a shorter banner-read window separate from the dial timeout, would keep it blazing.

minor: the ^-anchored banner regexes miss servers that emit a preamble before the ident line (ssh allows it per 4253), since Go ^ is start-of-text. fine to leave.

rest is solid. decodeTCPData bounds check out including the short/malformed \x cases, and validateTCP rejecting status/favicon at load matches what the executor will actually run. modules are low-FP. approving.

TBX3D:
rebased. this branch was carrying the whole tcp executor from before #308
landed, so it was nine commits against a base that has since merged; it is now
main plus the two commits that are actually this fix. that also dropped a stale
regression where the old base reverted the m.CaseInsensitive argument
checkTCPMatcher now passes to checkWords.

retitled to match what is left: "fix(tcp): stop a cancelled context from being
swallowed by the deadline re-arm".

on your follow-up note, agreed that readTCP still drains to the read deadline so
a peer that banners then holds the connection open costs the full timeout. happy
to take that as a separate change.

PR #383, opened 2026-07-31
fix(modules): validate modules at load instead of at match time

TBX3D:
stacked on #382, so only the commits above it are this pr's.

six more guards over the same theme, a module that loads clean and then quietly does nothing. unknown matcher types were let through where dns and tcp already reject them, regex compiled lazily at match time and swallowed the error, a type: http module with no http: block only failed at Execute, which a passive scan never reaches, and a typo'd severity ranked against nothing downstream.

the loader itself had two. loadDir aborted the whole walk on one unreadable entry, and LoadAll skipped the embedded set wholesale whenever anything loaded from an on-disk modules/ dir, so running sif from a directory that happens to contain one replaced every builtin. embedded is now the baseline that disk layers over by id.

### DIFF
--- a/internal/modules/executor.go
+++ b/internal/modules/executor.go
@@ -642,6 +642,24 @@ func checkRegex(content string, patterns []string, condition string) bool {
 	return true
 }
 
+// validateExtractors rejects a regex extractor whose pattern fails to
+// compile, at load rather than at match time: runExtractors (and its dns/tcp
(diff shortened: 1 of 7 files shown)


vmfunc:
the embedded-vs-disk loader fix here is genuinely worth it - that's a real bug. but this re-includes #382 (just merged) and rewrites the same validateMatchers/executor bits the dsl work touches. rebase onto main and it'll shrink to just the loader validation - then i'll take it.

---

PR #384, opened 2026-07-31
feat(modules): add a dsl matcher over response variables

TBX3D:
stacked on #379, so only the commits above it are this pr's.

modules can match on words, regex, status and size, but not on a relation between them, so anything like "200 and body contains x and it took under 2s" needs a new matcher type each time. this adds a dsl matcher over the nuclei expression language with a curated helper allowlist, compiled at load so bad syntax, a non-allowlisted function or an over-length expression fails before scan time rather than silently missing. host is bound to host[:port] from the parsed url, not the full url, so a nuclei-style host == "..." behaves the way someone pasting it expects.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -311,6 +311,48 @@ both 32-bit forms are accepted, so values from shodan or any favicon-hash tool
 drop in without conversion. pair it with a `status: 200` matcher so an error
 page served for `/favicon.ico` is not hashed. a finding fires when the body
 hashes to any listed value.
+### dsl matcher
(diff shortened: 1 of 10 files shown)


vmfunc:
the dsl matcher design is great - allowlisted helpers, the byte cap, fail-closed eval, compile-at-load. but it's stacked on #379 (just landed) so it's carrying that refactor too. rebase onto main so the diff is just the dsl matcher and it's in. this is the base for 385/386/388 so it goes next.

---

PR #385, opened 2026-07-31
feat(modules): bind nuclei url-part vars in dsl matchers

TBX3D:
stacked on #384, so only the commits above it are this pr's.

a pasted nuclei expression referencing baseurl or rooturl or tld hits an undefined variable, which is now a load-time failure rather than a silent miss, so this binds the capitalized url-part set and the dns vars alongside the response ones. it delegates to nuclei's own utils.GenerateVariables rather than reimplementing them, because the semantics are deliberately counterintuitive, host is the hostname without port while hostname carries it, path is the directory, port defaults by scheme, and a hand copy would drift on every nuclei bump. nuclei is already a direct dependency, so go.mod and go.sum are unchanged.


vmfunc:
delegating to GenerateVariables is the right call - won't drift on a nuclei bump. holding til #384 is in; rebase on top once it lands and this drops to just the url-part diff.

---

PR #386, opened 2026-07-31
feat(modules): named payload sets with streaming generation

TBX3D:
stacked on #384, so only the commits above it are this pr's.

the payloads field only took one anonymous list, so a fuzzing module could vary exactly one thing, and generation materialized the whole cross product before sending anything.

a mapping now declares ordered named sets injectable at {{name}}, a scalar names a local wordlist loaded through the existing loadWordlist, and a sequence still means the legacy single set. generation becomes a lazy iter.Seq consumed by a fixed worker pool over an unbuffered feed with one collector owning the findings slice, so the mutex and the product-sized results channel both go away. -fuzz-max-requests caps a module at 25000 requests per target by default.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -311,6 +311,48 @@ both 32-bit forms are accepted, so values from shodan or any favicon-hash tool
 drop in without conversion. pair it with a `status: 200` matcher so an error
 page served for `/favicon.ico` is not hashed. a finding fires when the body
 hashes to any listed value.
+### dsl matcher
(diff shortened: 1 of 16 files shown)


vmfunc:
the streaming rewrite is the right move - dropping the product-sized channel especially. it sits on #384 and collides with the ssl branch on the same executor refactor. once #384 lands, rebase and i'll take this + #387 as a pair.

---

PR #387, opened 2026-07-31
feat(modules): implement batteringram and a scan-wide fuzz budget

TBX3D:
stacked on #386, so only the commits above it are this pr's.

validateAttack rejected batteringram alongside sniper as not implemented. it is the simple one: every payload position in a request takes the same value at once, drawn from the first declared set, crossed with paths and stopping at that set's length rather than clusterbomb's product. it rides the existing lazy iterator, so nothing is materialized. the second commit adds the ceiling the per-module cap does not give: FuzzBudget is a shared atomic reserved per request by every producer, sized once per run by -fuzz-global-max-requests and reused across concurrently scanned targets, so exhaustion logs once for the scan instead of once per module.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -311,6 +311,48 @@ both 32-bit forms are accepted, so values from shodan or any favicon-hash tool
 drop in without conversion. pair it with a `status: 200` matcher so an error
 page served for `/favicon.ico` is not hashed. a finding fires when the body
 hashes to any listed value.
+### dsl matcher
+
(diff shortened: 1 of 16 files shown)


vmfunc:
batteringram and the shared budget both look right. this is the tip of the #384 → #386 stack though so it can't move til those do. one thing for the rebase: Reserve() bumps the counter even on the failed reservation, so the budget's off by one per module - tidy that up.

---

PR #388, opened 2026-07-31
feat(modules): add ssl module type

TBX3D:
stacked on #385, so only the commits above it are this pr's.

dns and tcp cover name resolution and raw sockets, but nothing in sif looks at the handshake itself, so an expired, self-signed or tls 1.0 endpoint reads as healthy.

ssl dials the port, handshakes with verification off since the module inspects the certificate rather than trusting it, and exposes tls_version, cipher, cn, san, not_after, self_signed and expired to the existing matcher, extractor and dsl machinery. self-signed detection uses CheckSignature and not CheckSignatureFrom, because the latter also enforces CA basic constraints, which a hand-rolled leaf usually omits and would then read as not self-signed. ships tls-certificate-health.yaml as a working example.


vmfunc:
good work - the self-signed rationale and the InsecureSkipVerify justification are exactly right, and ssl handshake inspection is a real gap. it's stacked on #385 → #384 though, so it waits on both. rebase after they land, and double-check the nil Resp path through dslVars/getPart stays clean for ssl.

---

